Issue #189, opened 2026-07-30
[Security] Live third-party API key committed in appsettings for dev, stg and prod: needs rotation, not deletion


iinception:
## What

`GeolocationApiKey` is present as a non-empty 32 character literal in all three environment configuration files:

- `server/Api/appsettings.dev.json`
- `server/Api/appsettings.stg.json`
- `server/Api/appsettings.prod.json`

It pairs with `GeolocationApiUrl = https://ipgeolocation.abstractapi.com/v1`, so it is a credential for a third-party geolocation service, and the production file carries a distinct value from the other two.

SonarQube reports this as `csharpsquid:S6418` on all three files. The value is deliberately not reproduced here.

## Why this needs a person, not a commit

**Deleting the key from the working tree does not fix anything.** It is already in git history, and this repository is heading for a public MIT release, so history will be published with it. Rewriting history to hide it is explicitly not an option under the project rules and would not help either, because the key has already been distributed to everyone who has ever cloned this repo.

The only thing that actually closes the exposure is **rotating the credential at abstractapi.com**, then keeping the new value out of the repository.

## Suggested sequence

1. Rotate the key at the provider. Treat the current values for all three environments as compromised.
2. Move the new values into the same vault mechanism the rest of the secrets already use, so the configuration files carry a placeholder rather than a literal.
3. Only then remove the literals from the configuration files. Doing this step first creates the false impression that the exposure is closed.
4. Confirm whether the dev and staging values are the same credential as production. If they are, rotation covers all three; if not, all three still need rotating.

## Related, lower severity, in the same sweep

`server/Api/Properties/launchSettings.json` contains `ASPNETCORE_Kestrel__Certificates__Default__Password` as a short literal, alongside a certificate path under an individual developer's home directory. That is a local development certificate rather than a production credential, so the impact is much smaller, but it is still a committed password and the file also leaks a developer's local filesystem layout. Worth cleaning up when the above is done. Reported by SonarQube as `csharpsquid:S2068`.

## What I did not do

I did not delete the key, mark the finding as safe, or touch git history. A hardening pass can fix code, but it cannot rotate a credential, and removing the literal without rotating would quietly convert a visible problem into an invisible one.

Raised during the HARDEN-9 pass on this repo.

iinception:
## Four more committed password literals found in the same sweep, needing a human eye

Separate from the geolocation key above, SonarQube reports `typescript:S2068` on four password literals in test fixture files:

[...]

PR #190, opened 2026-07-31
HARDEN-9: repair the dev merge, raise frontend coverage to 89.87%, clear 7 hotspots

iinception:
### Before and after, every number measured in this pass

| Gate | Before | After |
|---|---|---|
| Backend unit coverage | 89.84% (20487/22804) | 89.84%, held |
| Backend tests | 2002 + 1 skipped, green | unchanged, green |
| Frontend unit coverage | 84.83% (6328/7459) | **89.87%** (6704/7459) |
| Frontend tests | 1611 in 267 files | **1804** in 279 files, green |
| SonarQube bugs (inception) | 0 | 0 |
| SonarQube vulnerabilities | 0 | 0 |
| SonarQube hotspots TO_REVIEW | 11 | **4** |
| Dependency-Track vulnerabilities | 0 | 0 |

Frontend was the ratchet this pass, from 84.83% to 89.87% against a 89.83%
target. Backend was already above the hold line and stayed exactly where it was.

### The dev merge did not compile, and said it did

Merging dev brought fourteen payment commits that reshape provider encryption
around a per-organization key ring. Git merged them with no textual conflict,
but the result did not build: tests added on inception still referenced
interfaces dev had replaced.

This is worth calling out because of how it presented. `dotnet test --no-build`
runs the previous assembly when the build fails, so the suite reported 1885
passing at the old 90.68% while nothing had actually been compiled. The first
honest number after a successful build was 89.84%.

What moved, and how each test follows it:

- `IAesGcmSecretProtector.TryProtect` became `ProtectAsync`, returning a
  `SecretProtectionResult` and scoped by `PaymentEncryptionScope`.
- `IProviderCredentialRotationStrategy.CreatePlan` became `CreatePlanAsync`.
- `IPaymentProviderCache.Remove` and `RefreshAsync` are keyed by provider name
  as well as tenant and organization.
- `PaymentProvidersController` gained an `IPaymentEncryptionAdminService`.

Only mock plumbing moved; every assertion is unchanged.

One test was removed rather than adapted.
`Secret_readiness_defaults_to_available_and_is_not_overwritten` asserted that a
`PaymentSecretReadiness` singleton survives `TryAddSingleton`. Dev deleted that
record, because per-provider token-encryption readiness is now a
per-organization key ring, so there is nothing equivalent left to assert. The
concern is covered in more detail by dev's own
`ProviderTokenEncryptionKeyRingProviderTests`, which pin a missing ring failing
only its own organization and the health check reporting an unreadable ring. The
sibling assertion in the same file now names the successor hosted service,
`PaymentConfigurationReadinessLogger`.

### Frontend coverage

The payment UI that dev's backend work exists to serve had no tests at all:
eleven files at zero percent, and together most of the frontend gap. Twelve new
suites cover the list page's cursor paging and draft-then-apply filtering, the
provider create, update, rotate and list pages, the saved-methods section and
table, the filter panel and multi-select, and the stored-methods hook.

The one worth singling out is the refund dialog's idempotency key: the same key
is reused when an unchanged request is retried, and a fresh one is issued once
the amount changes, so a retry cannot refund twice. That is now pinned.

### SonarQube hotspots, 11 down to 4

Fixed in code: two regexes flagged for super-linear backtracking. In
`magic-url/utils/url.util.ts` the host group already matched dots, so letting the
suffix group match them too left the two ambiguous; the suffix is now letters
only, and a multi-label suffix still parses because the greedy host group absorbs
everything up to the last dot. In `oidc-utils.ts` a trailing-wildcard replace
became a split.

Marked safe with a written justification: four test-fixture passwords, after
first making them unmistakably synthetic, having verified that none of the
literals appear anywhere outside the fixture directories and none are asserted
on; and the `Math.random` in `getUniqueID`, which builds a display reference code
rather than anything relied on for authorisation.

### Deletions

Twelve lines: the commented-out five-project limit rule in
`CreateProjectRequestValidator` and `HasNotExceededProjectLimit`, which existed
only to serve it.

Deliberately kept: the `MongoQueryHelper` example-input comments, which document
query shapes rather than being disabled code, and the e2e consent block in
`login.spec.ts`, whose own comment says to confirm the button label with codegen
and enable it if the screen appears.

### Not done

The remaining 4 hotspots all need a human decision rather than a code change.
Three are the `GeolocationApiKey` in the dev, stg and prod appsettings, still
pending rotation under #189; deleting them would leave the value in history,
which is worse in a repository that is about to be published. The fourth is a
local development certificate password in `launchSettings.json`, added to #189
with the reasoning.


iinception:
### Sanity pass 3 (blocks-utilities)

All numbers below were produced against the pushed head `7e53d37`, not carried over from an earlier run.

**Both coverage gates are now cleared.** Backend and frontend were each sitting just under 90% and both are now above it.

| Target | Baseline this fire | After | Verdict |
|---|---|---|---|
| SonarQube bugs | 0 | 0 | held |
| SonarQube vulnerabilities | 0 | 0 | held |
| Hotspots TO_REVIEW | 4 | 4 | blocked, see #189 |
| Dependency-Track vulns | 0 | 0 | held, fresh SBOM |
| Backend coverage | 89.84% (20495/22812) | **90.36%** (20614/22812) | over the gate |
| Frontend coverage | 89.87% (6704/7459) | **90.27%** (6724/7448) | over the gate |
| Backend tests | 2003 pass, 1 skip | 2014 pass, 1 skip | +11 |
| Frontend tests | 1804 | 1817 | +13 |

#### The dev merge

Merged clean this time, 19 files, all server side, no lockfile moved. It brings the refund and capture work: `PaymentRefundRequestFactory`, `PaymentRefundInitiationService`, `StoredPaymentMethodQueryService`, `StripeRefundProviderGateway` and the Stripe routing metadata. The build was checked for `error CS` before any test result was trusted, which is what the last pass on this repo got caught by.

#### Coverage, reported twice as required

Frontend after new tests and before deletions: **90.15%** (6713/7459). After deleting the three unused components: **90.27%** (6724/7448). The deletions account for 0.12pp of the movement and the tests for the rest. Backend moved on tests alone since no server file was deleted.

Three suites added, all asserting behaviour rather than touching lines:

- `PaymentEncryptionAdminService` was the largest untested unit left on the server, 0% of 90 lines. Its refusals are the interesting part. An unreadable ring is carried through `GetHealth` as a *successful* diagnostic, because "the ring cannot be read" is a valid answer to "is this ring healthy". And `ReEncrypt` refuses outright on the shared ring, since re-encrypting there would move nothing and report success, which reads as a completed migration. Also pinned: the scope is always derived from the caller's own context and never taken as a parameter, which is what stops one tenant rewriting another tenant's ciphertext under its own key.
- `useCreatePaymentRefund` arrived with the refund work untested. Its substance is the optimistic cache write that stops a second refund being submitted before the invalidated read returns. Covered: the flag lands on the right row and only that row, it is keyed by the selected project so it cannot mark a payment in another tenant's list, an empty cache slot is left alone rather than written as a partial list, and a refused refund leaves the cache untouched.
- `UserMFAConfirmationDisable` was at 8%. A server refusal comes back as a normal response rather than a rejection, so there are two distinct failure paths and closing on the wrong one would tell the user MFA is off while it is still on.

#### Deletions

108 lines, well inside the cap. All three were proven unreachable from `main.tsx` and the router, and named nowhere else by path, by exported symbol, through a barrel or through a dynamic import:

- `date-range-filter.tsx`, superseded by `filter-toolbar/date-range`. The only other `DateRangeFilter` match in the tree is a coincidental props type inside that replacement.
- `configure-button.tsx`, sitting beside a live `primary-button` with no callers left.
- `environment-card-skeleton.tsx`, sitting beside an `environment-card` that is used and tested but was never wired to it.

No commented-out code remains anywhere in `client/app` or `server`. The previous pass took all of it and the dev merge added none.

#### What I chose NOT to delete

Five further modules are equally unreferenced and were deliberately left:

- `routes/auth/resetpassword.tsx`, `guards/project-guard.tsx` and `back-to-console-navigator.tsx`. All three were last touched on 2026-07-29 by the same author and arrived as a group: a guard, an auth route and a navigator added together but not yet wired. That is in-flight work, not dead code. An auth route is also on the never-delete list.
- `project-overview-sidebar-mobile.tsx` and `project-overview-sidebar-desktop.tsx`, a matched pair of layout shells from 2026-07-13. Layout shells are entrypoint-shaped and the pair reads as pending wiring rather than abandonment.

#### Hotspots, still 4, still blocked on the team

Unchanged and correctly outstanding, all tracked in #189:

- Three `S6418` on the `GeolocationApiKey` in `appsettings.dev.json`, `appsettings.prod.json` and `appsettings.stg.json`. **The key is unchanged by this merge and still not rotated.** Worth restating: all three files carry the *same* key, so the leak in the dev config exposes production too. Nothing here can be fixed by an edit, because the value stays in history regardless. It needs rotation.
- One `S2068` on the Kestrel dev-certificate password in `launchSettings.json`, which also leaks a developer's home directory path. Moving it to user secrets is the team's call.

Marking any of these reviewed would be false while the key is live, so they stay TO_REVIEW.

#### E2E not run

The Playwright suite (2 spec files, 19 declared tests) cannot start in this environment. The web server it launches fails at boot with `Required Azure config value 'KeyVaultUrl' is missing`, so Playwright reports `Process from config.webServer was not able to start`. That is an infrastructure gap, not a regression, and nothing in this pass touches the startup path. Flagging it rather than reporting a green suite I did not see.

Backend 2014 passing, 1 skipped. Frontend 1817 passing across 281 files. Both builds clean.

PR #250, opened 2026-08-19
Make subscription catalogues client-ready

fahad0208:
## Summary

Makes the subscription catalogue usable by client applications and implements the seven agreed workstreams.

- removes the fictional pricing-shape choice, separates quantity items and meters, inherits entitlement limits, renames the grant step, and adds plan duplication
- adds configurable, snapshotted usage windows independent of fee cadence
- allows cross-interval plan changes and atomically rebuilds fee and usage schedules with proration
- adds plan-family metadata, family catalogue grouping, ranks, and display price notes
- implements a scoped discount catalogue with validation, retirement, subscription-time lookup, and immutable subscription snapshots
- makes subscription payments visible to subscriber organizations, attributes checkout payments consistently, and exposes payment flow/invoice availability
- updates the maintained subscription documentation, including price archival, discounts, invoices, and the first-checkout invoice boundary

## Impact

Clients can express flat-fee, per-seat, metered, mixed, family-tiered, and discounted catalogues without relying on inferred UI state. Subscription schedules now follow stored plan terms, cadence changes are safe, and subscriber payment lists can render invoice download actions without probing each payment.

## Verification

- `dotnet test server/XUnitTest/XUnitTest.csproj --filter "FullyQualifiedName!~Integration" --no-restore --verbosity minimal -p:WarningLevel=0` — 2,179 passed, 1 skipped
- subscription/payment focused server tests — 328 passed
- client TypeScript typecheck — passed
- client ESLint for changed subscription/router files — passed
- `vitest run app/cross-modules/utilities/subscription --maxWorkers=1` — 106 passed
- `git diff --check` — passed

The repository-wide client `npm test` command was attempted with default workers and one worker, but hangs after unrelated zero-test/jsdom canvas suites; the focused subscription suite is clean.

Integration tests were not run because they require a local MongoDB instance.

## Known repository boundary

This client package has no subscriber checkout/subscribe form. The server now validates and applies `discountCode`, and the client includes discount catalogue authoring; a checkout UI in its owning application can pass the existing request field.

### DIFF
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx
@@ -41,7 +41,7 @@ import { StepUsageLimits } from "./step-usage-limits";
 const STEPS: Steps = [
   { id: 1, title: "Identity" },
   { id: 2, title: "Pricing model" },
-  { id: 3, title: "Usage limits" },
+  { id: 3, title: "What the plan grants" },
   { id: 4, title: "Trial" },
   { id: 5, title: "Review" },
 ];
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
@@ -84,7 +84,10 @@ export const PlanPriceFields = ({
                 key={price.priceId}
                 className="flex items-center justify-between gap-3 text-sm"
               >
-                <span>{formatPrice(price)}</span>
+                <span>
+                  {formatPrice(price)}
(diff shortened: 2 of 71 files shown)


fahad0208:
Addressed in deece96. The plan-change CAS now atomically queues an immutable snapshot of the outgoing usage window before installing the new schedules; the usage-rating sweep prices that window with its original plan/price terms and removes it only after the idempotent invoice exists. This preserves accrued overage across crashes/concurrent changes and prevents an unbilled allowance reset. I also made target proration boundaries required and documented that cross-interval changes charge a full new target period immediately, less unused old-period value. Regression coverage was added for atomic handoff and old-allowance rating. Full non-integration suite: 2,181 passed, 1 skipped.

Issue #274, opened 2026-08-23
[Architecture] Durable root scheduler for subscription and payment background work


fahad0208:
## Problem

`SubscriptionReconciliationBackgroundService` refreshes the complete tenant roster and processes every tenant sequentially. With roughly 2,394 tenants, due subscription or payment work for a tenant late in the roster can wait behind thousands of tenants that have nothing to process.

The current sweep also performs several tenant-level checks per iteration (activation/recovery, settlement recovery, renewal, usage-period closure, invoice charging, and outbox publication). This creates avoidable database traffic, unpredictable financial-processing latency, and weak operational visibility into which work is due, leased, retried, completed, or dead-lettered.

Because these workflows move money, simply parallelizing the full roster scan is not a sufficient long-term design.

## Goal

Introduce a durable, centrally queryable work scheduler in `BlocksRootDb` for subscription- and payment-related background work. Workers should claim only work that is due, while tenant databases remain the authoritative source of subscription and payment state.

A slow full-tenant sweep must remain as a repair/reconciliation safety net, not the primary scheduler.

### Root collections

Prefer module-owned collections:

- `SubscriptionBackgroundWork`
- `PaymentBackgroundWork`

A shared `FinancialBackgroundWork` collection is acceptable only if ownership, deployment coupling, work-type versioning, and indexes remain clear.

Each work item should contain at least:

- `Id`
- `TenantId`
- `OrganizationId` where applicable
- `SubscriptionId` / `PaymentId` where applicable
- `WorkType`
- `WorkKey` identifying the occurrence (billing period, usage period, provider event, etc.)
- `Status`: `Pending`, `Processing`, `Completed`, `DeadLetter`
- `DueAtUtc`
- `NextAttemptAtUtc`
- `Priority`
- `AttemptCount` / `MaxAttempts`
- `LeaseId`, `LeasedBy`, `LeaseExpiresAtUtc`
- `CorrelationId` and `OperationId`
- sanitized `LastErrorCode` / `LastErrorMessage`
- `CreatedAtUtc`, `UpdatedAtUtc`, `CompletedAtUtc`
- `PurgeAtUtc` for terminal-record retention

Do not store card details, secrets, or complete provider/webhook payloads in these collections.

### Work types

Subscription work should cover, as applicable:

- pending checkout activation/recovery
- stale settlement-reservation recovery
- renewal
- usage-period rating/closure
- usage-invoice charging
- subscription outbox publication
[...]

PR #275, opened 2026-08-23
feat(subscription): durable work queue for subscription background work

fahad0208:
Refs #274. Implements steps 1–3 of the migration the ticket lays out, for **subscription** work. Read the scope section before reviewing — this is deliberately not the whole ticket.

## The problem, restated

The reconciliation sweep refreshed the whole tenant roster and processed tenants in order, running six checks against each. With ~2,400 tenants, a renewal due for a tenant late in the roster waited behind thousands with nothing to do. The wait was proportional to tenants that *exist*, not to work outstanding — and parallelising the scan would not fix that, because the scan is the cost.

## What this adds

| Piece | Role |
| --- | --- |
| `SubscriptionBackgroundWork` | One scheduled occurrence, in `BlocksRootDb`. No card data, no secrets, no provider payloads. |
| `ISubscriptionWorkQueue` | Schedule, claim, renew, complete, fail, list dead letters, report depth. |
| `ISubscriptionWorkScheduler` | Producer seam. Idempotent by occurrence; assigns priority. |
| `ISubscriptionWorkHandler` | Runs one kind of work. Thin by design — see below. |
| `SubscriptionWorkDispatcher` | Claims a bounded batch, runs it with bounded parallelism, records the outcome. |
| `SubscriptionWorkSchedulerBackgroundService` | The worker loop. |

A claim is one `FindOneAndUpdate` against one indexed collection, ordered by priority then by how overdue the item is, held by a lease another worker can take once it expires.

**Priority is money before bookkeeping.** Settlement recovery first — until it resolves, a subscriber may have paid for units they do not have, and their subscription can neither renew nor change plan — then activation, renewal, usage, and the outbox last.

## The consistency boundary

The two databases share no transaction, and the design says so rather than pretending otherwise:

| Failure | What heals it |
| --- | --- |
| Tenant state committed, scheduling write lost | The repair sweep finds it |
| Scheduling write committed, tenant state moved on | The handler re-reads and finds nothing to do |
| Worker died after the provider succeeded | Lease expires, item is reclaimed, and the handler's provider idempotency key — derived from persisted identity — finds the existing charge |

That last row is why **handlers are deliberately thin**. A renewal keys its charge on period + attempt; a settlement keys it on its reservation id. Those keys already exist and already make retries safe. Reimplementing them here would give the same money two sets of rules.

## Rollout

`Subscription:SchedulerEnabled`, **off** by default, read **once at startup**.

- **Off** — the sweep executes work exactly as today. Nothing changes.
- **On** — the sweep stops executing and starts *scheduling*; the scheduler runs the work.

Never both: executing in one and scheduling in the other would run the same work twice, and twice is a second charge. Reading the flag per pass would let the two disagree mid-flight, which is why it isn't.

## Verification

- Build clean; **2317 non-integration tests passing** (up from 2308)
- **9 dispatcher unit tests**: completion under the claiming lease, tenant context established before any handler reads, transient → requeue with backoff, permanent → dead-letter, handler exception → retry, unknown work type → dead-letter, whole batch runs, batch size and lease come from config
- **13 queue integration tests**: duplicate occurrence creates one item, eight concurrent claims yield exactly one winner, a live lease blocks reclaim, an expired one allows it, not-yet-due is left alone, priority beats age, completion sets `PurgeAtUtc` and dead-lettering does not, an attempt whose lease moved on can neither complete nor fail, backoff blocks reclaim until it passes, attempts exhaust into dead-letter, lease renewal holds an item, depth reports oldest due age

**The integration tests have not been run.** There is no mongod or Docker daemon on this machine, and they need one (`BLOCKS_IT_MONGO` points the fixture at a remote). They follow the existing `MongoIntegrationFixture` convention, which the payment integration tests already use and which is likewise excluded from the default filter — so CI or any machine with mongod will run them. Everything they assert is a property of the database rather than of a class, which is exactly why I did not fake them, and exactly why I cannot vouch for them yet.

## Scope — what is *not* here

- **Payment work types** (reconciliation, webhook recovery, provider refresh, cleanup). They belong in `PaymentBackgroundWork` with their own producers; this PR is subscription-only.
- **Per-aggregate producers at the point of state change.** The repair sweep is currently the only producer, so it still walks the roster — it just schedules instead of executing. The entity already carries `AggregateId` for when a renewal schedules its own next occurrence.
- **An operator endpoint for requeueing dead letters.** They are queryable and alert; requeueing is manual.

Consequently the acceptance criterion *"a due item for any tenant can be claimed without scanning tenants that have no due work"* holds for the **claim** — which is the latency that hurt — but production of that work still involves a roster pass until the inline producers land. Worth being explicit about rather than ticking.

### DIFF
--- a/.github/workflows/subscription-scheduling-integration.yml
+++ b/.github/workflows/subscription-scheduling-integration.yml
@@ -1,0 +1,46 @@
+name: Subscription scheduling integration
+
+on:
+  pull_request:
+    branches: [dev]
+    paths:
+      - "server/Subscription.DomainService/Scheduling/**"
+      - "server/Subscription.DomainService/Services/SubscriptionRenewalService.cs"
+      - "server/XUnitTest/Integration/**"
+      - ".github/workflows/subscription-scheduling-integration.yml"
+  push:
+    branches: [dev]
+    paths:
+      - "server/Subscription.DomainService/Scheduling/**"
+      - "server/Subscription.DomainService/Services/SubscriptionRenewalService.cs"
+      - "server/XUnitTest/Integration/**"
+
+jobs:
+  mongo-integration:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    services:
+      mongodb:
+        image: mongo:8.0
+        ports:
+          - 27017:27017
+        options: >-
+          --health-cmd "mongosh --quiet --eval 'db.adminCommand({ ping: 1 })'"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 12
+    env:
+      BLOCKS_IT_MONGO: mongodb://localhost:27017
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-dotnet@v4
+        with:
+          dotnet-version: 10.0.x
+      - name: Restore
+        run: dotnet restore server/XUnitTest/XUnitTest.csproj
+      - name: Run subscription Mongo integration suite
+        run: >-
+          dotnet test server/XUnitTest/XUnitTest.csproj
+          --no-restore
+          --filter "FullyQualifiedName~SubscriptionWorkQueueIntegrationTests|FullyQualifiedName~SubscriptionRenewalCrashRecoveryIntegrationTests"
+          --verbosity minimal
--- a/.github/workflows/subscription-scheduling-integration.yml
+++ b/.github/workflows/subscription-scheduling-integration.yml
@@ -2,7 +2,6 @@ name: Subscription scheduling integration
 
 on:
   pull_request:
-    branches: [dev]
     paths:
       - "server/Subscription.DomainService/Scheduling/**"
       - "server/Subscription.DomainService/Services/SubscriptionRenewalService.cs"
--- a/monitoring/subscription-background-work-alerts.yaml
+++ b/monitoring/subscription-background-work-alerts.yaml
@@ -1,0 +1,34 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: blocks-subscription-background-work
+spec:
+  groups:
+    - name: subscription-background-work
+      rules:
+        - alert: SubscriptionBackgroundWorkDeadLettered
+          expr: increase(subscription_work_dead_lettered_total[5m]) > 0
+          for: 0m
+          labels:
+            severity: critical
+            financial: "true"
+          annotations:
+            summary: Subscription financial work was dead-lettered
+            description: "{{ $labels.work_type }} produced a dead letter and requires operator review."
+        - alert: SubscriptionRenewalWorkLate
+          expr: subscription_work_oldest_due_age_seconds{work_type=~"Renewal|ActivationSettlement|UsageInvoiceCharge"} > 300
+          for: 5m
+          labels:
+            severity: critical
+            financial: "true"
+          annotations:
+            summary: Money-moving subscription work is late
+            description: "Oldest {{ $labels.work_type }} item has exceeded the 5 minute service threshold."
+        - alert: SubscriptionBookkeepingWorkLate
+          expr: subscription_work_oldest_due_age_seconds{work_type=~"ActivationRecovery|SettlementReservationRecovery|UsagePeriodClosure|OutboxPublication"} > 1800
+          for: 10m
+          labels:
+            severity: warning
+          annotations:
+            summary: Subscription background work is not draining
+            description: "Oldest {{ $labels.work_type }} item has exceeded the 30 minute threshold."
--- a/monitoring/subscription-background-work-dashboard.json
+++ b/monitoring/subscription-background-work-dashboard.json
@@ -1,0 +1,31 @@
+{
+  "title": "Subscription background work",
+  "uid": "blocks-subscription-work",
+  "schemaVersion": 39,
+  "panels": [
+    {
+      "type": "timeseries",
+      "title": "Queue depth by work type and status",
+      "targets": [{ "expr": "subscription_work_queue_depth", "legendFormat": "{{work_type}} / {{status}}" }]
+    },
+    {
+      "type": "timeseries",
+      "title": "Oldest due age (seconds)",
+      "targets": [{ "expr": "subscription_work_oldest_due_age_seconds", "legendFormat": "{{work_type}}" }]
+    },
+    {
+      "type": "timeseries",
+      "title": "Completed, retried and dead-lettered",
+      "targets": [
+        { "expr": "rate(subscription_work_completed_total[5m])", "legendFormat": "completed {{work_type}}" },
+        { "expr": "rate(subscription_work_retried_total[5m])", "legendFormat": "retried {{work_type}}" },
+        { "expr": "increase(subscription_work_dead_lettered_total[5m])", "legendFormat": "dead {{work_type}}" }
+      ]
+    },
+    {
+      "type": "timeseries",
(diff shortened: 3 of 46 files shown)


fahad0208:
Review against #274 — **blocking findings**:

1. **[P1] The rollout modes can diverge after a configuration reload.** `SubscriptionWorkSchedulerBackgroundService` reads `SchedulerEnabled` once and exits or keeps draining, but `SubscriptionReconciliationBackgroundService` reads `_options.CurrentValue.SchedulerEnabled` for every tenant/pass. If the process starts disabled and the setting becomes enabled, the sweep starts enqueueing while the scheduler has already exited, stranding every item. If it starts enabled and becomes disabled, the scheduler keeps draining while the sweep resumes direct execution, violating the stated “never both” financial-safety invariant. Capture one immutable startup mode in a shared singleton (or otherwise ensure both services use the same snapshot), and test both reload directions.

2. **[P1] Index-creation failure must fail closed.** `EnsureIndexesAsync` catches failure of the unique occurrence index and then starts the dispatcher anyway. That explicitly permits duplicate financial work items to be produced and drained. The unique index is described as a correctness boundary, so scheduler startup must not claim/process work until required indexes are confirmed. Retry index initialization with alerting, or stop the hosted service/process; logging and continuing is unsafe.

3. **[P1] Claimed work has no lease heartbeat.** `RenewLeaseAsync` exists but is never called. Any handler exceeding `SchedulerLeaseSeconds` can be reclaimed and run concurrently by another worker. The first attempt then ignores the boolean from `CompleteAsync` and logs completion even if its lease was lost. Add per-item lease renewal while the handler runs, stop treating a lost completion CAS as success, and add a test with a handler running beyond the original lease.

The queue primitives and dispatcher structure are otherwise a solid first subscription-only slice. Local verification: Worker build passed; 9/9 `SubscriptionWorkDispatcherTests` passed. The 13 Mongo integration tests were not run here, so their claims remain unverified.

This PR is intentionally partial relative to #274. Payment work, point-of-change producers (so roster scanning is no longer the primary producer), actual metrics/alerts, cross-API correlation, operator dead-letter recovery, and the provider-success/worker-crash integration test remain open.

fahad0208:
Re-review of latest head `e6184e4954846ed9b15f0bbc0d2802726a9f179d` against #274.

The previous in-process reload issue is fixed: both hosted services now share one frozen `SubscriptionSchedulerMode`. Completion CAS failures are no longer reported as success, lease renewal is now exercised, and the focused dispatcher suite passes 12/12 locally.

Two blocking correctness gaps remain, plus one rollout requirement that must be made explicit:

1. **[P1] Producers still run before the correctness indexes exist.** `WaitForIndexesAsync` gates only the scheduler consumer. In queue mode, `SubscriptionReconciliationBackgroundService` starts scheduling immediately and `SubscriptionWorkQueue.ScheduleAsync` uses `InsertOneAsync`. If two sweeps/processes produce the same occurrence before the unique index exists, duplicates can be inserted. Index creation then fails on those duplicates forever, while the scheduler correctly refuses to drain: the queue is wedged and needs manual deduplication. Gate scheduling and claiming behind the same shared “indexes ready” state, or initialize indexes before either hosted service can produce/consume.

2. **[P1] Renewal exceptions do not fail safe at lease expiry.** The heartbeat catches any exception from `RenewLeaseAsync`, logs it, and keeps the handler running. If root Mongo is unavailable for longer than the remaining lease, another worker can reclaim the item as soon as Mongo recovers while the first handler is still moving money. A failed renewal is not proof the lease is lost, but once the locally known expiry is reached without a confirmed extension, ownership is no longer defensible. Track the confirmed lease deadline; cancel the handler before that deadline unless renewal succeeds. Add a test where renewal throws through expiry and verify the handler is cancelled and records neither completion nor failure.

3. **[P1 rollout] “Never both” is process-local, not deployment-wide.** During a normal rolling deployment that changes `SchedulerEnabled`, old pods can retain direct mode while new pods start queue mode. The shared singleton prevents disagreement inside one process but not across worker replicas. Either coordinate the mode in BlocksRootDb, use a deployment strategy that guarantees zero overlap when this flag changes, or document and enforce a two-phase rollout that cannot run both generations together. Add a multi-instance/rollout test or operational guard.

Local verification:
- `SubscriptionWorkDispatcherTests`: 12 passed, 0 failed.
- Worker/API dependencies built as part of the test run.
- Mongo integration tests remain unexecuted locally; database-level lease, unique-index, TTL, and concurrent-claim criteria remain unverified.

PR #275 remains a useful subscription-only foundation, but issue #274 is still intentionally incomplete: inline producers, payment work, real metrics/alerts, end-to-end tracing, operator recovery, and provider-success/crash verification are outside this PR.

fahad0208:
Re-review of head 55d84b51d37b2da4933c2791ef0a6a1bf6627566 against #274.

The prior producer/index race is fixed: ScheduleAsync and ClaimDueAsync now both await the queue-owned index initializer, and failed initialization is not cached. Exception-based lease-renewal failure is also bounded by the last confirmed lease, with a focused test. The dispatcher suite passes 13/13 locally.

One lease-safety gap remains:

**[P1] A hung renewal call can still outlive the confirmed lease.** The heartbeat awaits RenewLeaseAsync with CancellationToken.None. The new confirmedUntil check only runs after that await throws or returns. If the Mongo operation stalls longer than the remaining lease, the loop cannot perform the deadline check, the handler keeps running, and another worker may reclaim the item after Mongo recovers. Bound each renewal call by the remaining confirmed lease (preferably with safety margin), pass a cancellable token, and independently cancel the handler when that deadline is reached. Add a test where RenewLeaseAsync never completes rather than throwing.

The rolling-deployment mismatch is now documented, but not enforced. It remains an open issue criterion: a normal rolling restart can still overlap direct-mode and queue-mode replicas. The documented full-restart procedure is operational guidance, not a technical guard.

Verification:
- SubscriptionWorkDispatcherTests: 13 passed, 0 failed.
- Worker/API dependencies built during the test run.
- Mongo integration tests remain unexecuted locally, so atomic multi-worker claims, expired-lease reclaim, unique occurrence behavior, and TTL behavior remain unchecked.

This is still a partial subscription-only slice of #274. Payment work, point-of-change producers, metrics/alerts, end-to-end trace continuity, operator recovery, and provider-success/crash integration coverage remain outstanding.

fahad0208:
Re-review of head 7173bf5beef216e73a935265164b3dd5468f11dc against #274: no remaining code blocker in the subscription-queue slice.

The hung-renewal finding is resolved correctly:
- the Mongo renewal call receives a deadline-bound cancellation token;
- an independent watchdog does not depend on the call observing cancellation;
- confirmed ownership advances only after a true renewal result;
- deadline loss cancels the handler and records neither completion nor failure;
- abandoned late-faulting tasks are observed.

Local verification:
- SubscriptionWorkDispatcherTests: 15 passed, 0 failed.
- Worker/API dependencies built successfully during the test run.
- CodeQL and repository-hygiene checks are green.
- Mongo integration tests remain unexecuted locally, so their database-level assertions are not independently verified.

The activation runbook and warning-level mode logs are adequate for this PR's agreed rollout boundary, provided SchedulerEnabled remains off on merge and activation uses a full fleet stop/start. Fleet-wide distributed mode coordination is still an open #274 item, not implemented here.

PR #275 is ready as the subscription queue foundation, but it does not close #274. Remaining issue scope includes payment work, point-of-change producers (the roster sweep is still primary production), production metrics/alerts, end-to-end trace continuity, operator dead-letter recovery, the provider-success/worker-crash integration test, and execution of the Mongo integration suite.

PR #306, opened 2026-08-25
feat(subscription): automatic discounts per price

fahad0208:
Adds an automatic percentage discount to every plan price, so a monthly and a yearly price under one plan can be discounted differently without a code.

## What this is

A price can now reduce itself:

```json
{ "automaticDiscountBasisPoints": 800, "quantityDiscountCombination": "Additive" }
```

Pro Monthly carries nothing, Pro Yearly carries 8%, and both stay prices on one plan rather than two products. It lives on the **price**, not the plan, because the offer is cadence-specific — and because a plan sold in two currencies has two prices that may not be discounted alike.

## Two combinations, because there are two questions

Three reductions can now land on one charge, and the interesting design decision is that they are not one negotiation:

- **`BuiltInDiscountCalculator`** settles the two the *merchant* authored — the price's own discount and the volume band the quantity selects — the way the **price** says. `BestDiscount` takes whichever reduces the charge by more money and only that one; `Additive` adds the rates and applies the sum once, so 8% + 5% is **13%**, not the 12.6% that applying one after the other gives. Capped at 100%, because no charge may arrive negative.
- The plan's existing **`QuantityDiscountCombinationPolicy`** then settles what a *redeemed code* adds to that result, exactly as it already settled band-versus-code. `QuantityOnly` now means "built-in discounts only" — the stored wire value is unchanged, so an existing plan means what it always did.

Collapsing the two would make "8% for paying yearly" negotiate with a coupon, which is not a question anybody authoring a catalogue is asking.

A promotion that loses is still not consumed, and that now includes losing to an automatic discount: three months of "2% off" are not spent on periods where the price's own 8% was larger.

## Order of operations

**gross → built-in discount → promotional code → tax → credit.** Tax after the discounts because tax is owed on what is actually charged; credit last because a credit pays a bill without changing what the bill was for. One calculator serves checkout, renewal, quantity preview, plan-change proration, simulation and usage overage, so no two paths can disagree about the same charge.

Discounts **truncate** to the minor unit — the direction that favours the subscriber, and the one the existing volume bands already took. A price with **no** automatic discount goes through the new calculator and comes out with its band's own arithmetic untouched to the minor unit, which is the state every stored price is in. There is no repricing of anything already sold.

## Snapshots, and what an edit can reach

Both fields are snapshotted at signup and at plan change, like the amount and the tax. `PUT /api/subscription-plans/prices/{priceId}/discount` sets or clears the discount for **future subscriptions and future moves onto the price**; nobody already on it is repriced in either direction. A plan change snapshots the *target* price's discount and prorates both sides at their own.

Metered overage is charged by the price the subscription was sold on, so the automatic discount reaches an overage invoice too, before its tax. A volume band does not — it prices units of a quantity item and a meter has none — and a promotional code still does not reach usage invoices at all. That asymmetry is deliberate and documented rather than incidental.

## Promotional codes gain price applicability

`applicablePriceIds` beside the existing `applicablePlanCodes`. The two lists **narrow** rather than offering two ways to qualify: with both set, both have to match, so a code aimed at the yearly price is refused on the monthly one with `subscription_discount_not_applicable`. Discounts stored before this carry no price list and stay unrestricted by price.

## What a caller can now see

Subscription responses, quantity previews and overage invoices report the automatic rate, the combination, the gross, the built-in reduction, the promotional reduction and the amount that was then taxed — reported rather than left to be reverse-engineered, since recovering a discount from a total means knowing which combination produced it. Signup writes one financial log line naming the price, the rates, the policies and the reductions, with identifiers hashed the way the rest of this module hashes them.

Provider invoice lines continue to show the discounted subtotal rather than a separate discount line: the lines have to add up to what was charged, and a gross line plus a reduction line is a different invoice shape than this module writes.

## UI

Each price card in the plan builder gains **Automatic discount (%)** and **Combine with quantity discount**, plus a preview that prices the *default quantity* through the band that quantity selects and then taxes what is left. The preview is the point — an author choosing between "use the better discount" and "add the percentages" cannot otherwise see that the two answers differ by real money, and finding out on a customer's first invoice is the expensive way to learn which one they picked. Existing prices get an editor beside the tax one, worded so it is clear it reaches future subscriptions only.

The discount catalogue replaces its comma-separated plan-code box with plan and price selectors, the price list filtered by the plans chosen, and each catalogue row states what it applies to in words rather than identifiers.

## Verification

- Server unit suite **2526 passing**. The 4 `SubscriptionSimulation*` permission-guard failures are pre-existing on `dev` and untouched.
- New: `AutomaticPriceDiscountTests` (24) covering both combinations, the 100% cap, the untouched no-discount path, all three promotion policies, tax ordering in both modes, credit last, and a reconciliation property — gross less both reductions is what was taxed, plus tax less credit is what was charged. Plus price applicability and snapshotting in the creation, plan-change, catalogue and usage-rating suites.
- Client: typecheck clean, production build succeeds, `app/cross-modules` **827 passing**. Six unrelated test files fail to transform an SVG asset — verified failing on `dev` too.

## Open question

Metered overage now carries the price's automatic discount. The spec said "usage charges where applicable" and the assumption that all charges associated with the price use its configuration, so this reads as intended — but it is the one place where a percentage authored for a recurring fee reaches a usage-rated line, and it is easy to reverse if that is not wanted.

### DIFF
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx
@@ -71,6 +71,7 @@ export interface PlanBuilderProps {
    */
   onRetirePrice?: (priceId: string) => void;
   onUpdatePriceTax?: (priceId: string, taxPercent?: number, taxMode?: "Exclusive" | "Inclusive") => Promise<void>;
+  onUpdatePriceDiscount?: (priceId: string, discountPercent?: number, combination?: "BestDiscount" | "Additive") => Promise<void>;
   retiringPriceId?: string | null;
   /** Rejecting leaves the draft alone and shows the reason; the caller navigates on success. */
   onSubmit: (values: CreateSubscriptionPlanFormValues) => Promise<void>;
@@ -94,6 +95,7 @@ const PlanBuilderWizard = ({
   existingPrices = [],
   onRetirePrice,
   onUpdatePriceTax,
+  onUpdatePriceDiscount,
   retiringPriceId = null,
   onSubmit,
 }: PlanBuilderProps) => {
@@ -226,6 +228,7 @@ const PlanBuilderWizard = ({
                       existingPrices={existingPrices}
                       onRetirePrice={onRetirePrice}
                       onUpdatePriceTax={onUpdatePriceTax}
+                      onUpdatePriceDiscount={onUpdatePriceDiscount}
                       retiringPriceId={retiringPriceId}
                     />
                   )}
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx
@@ -178,6 +178,11 @@ const PlanBuilderWizard = ({
           : price.quantityItemKey,
       taxRateBasisPoints: price?.taxPercent ? toBasisPoints(price.taxPercent) : null,
       taxMode: price?.taxPercent ? price.taxMode : null,
+      // The review step summarises what will be sold, and a price with 8% off it is not the price
+      // the amount alone says.
+      automaticDiscountBasisPoints: price?.automaticDiscountPercent
+        ? toBasisPoints(price.automaticDiscountPercent)
+        : null,
     })),
   };
 
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.test.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.test.tsx
@@ -173,6 +173,79 @@ describe("PlanPriceFields", () => {
     expect(screen.getByText(/\+ 20% tax/)).toBeInTheDocument();
   });
 
+  it("shows what an automatic discount does to the charge", async () => {
+    // The whole reason the preview exists: 8% off 89.00 is a number an author can check against the
+    // pricing page they wrote, and a percentage on its own is not.
+    render(
+      <Harness>
+        <PlanPriceFields />
+      </Harness>,
+    );
+
+    await userEvent.type(screen.getByLabelText(/^Amount$/), "89");
+    await userEvent.type(screen.getByLabelText(/Automatic discount for price 1/), "8");
+
+    const preview = screen.getByTestId("discount-preview-0");
+
+    expect(preview).toHaveTextContent("$7.12");
+    expect(preview).toHaveTextContent("$81.88");
+  });
+
+  it("offers the combination only once there is a discount to combine", async () => {
+    render(
+      <Harness>
+        <PlanPriceFields />
+      </Harness>,
+    );
+
+    expect(
+      screen.queryByLabelText(/Discount combination for price 1/),
+    ).not.toBeInTheDocument();
+
+    await userEvent.type(screen.getByLabelText(/Automatic discount for price 1/), "8");
+
+    expect(screen.getByLabelText(/Discount combination for price 1/)).toBeInTheDocument();
+  });
+
+  it("holds no discount at all when the field is cleared", async () => {
+    // The same trap the tax rate walks into: an emptied number input holds "", which coerces to 0,
+    // which authors a discount of nothing rather than no discount.
+    let latest: CreateSubscriptionPlanFormValues | undefined;
+
+    render(
+      <Harness onValues={(values) => { latest = values; }}>
+        <PlanPriceFields />
+      </Harness>,
+    );
+
+    const discount = screen.getByLabelText(/Automatic discount for price 1/);
+
+    await userEvent.type(discount, "8");
+    await userEvent.clear(discount);
+
+    expect(latest?.prices[0]?.automaticDiscountPercent).toBeFalsy();
+    expect(screen.queryByTestId("discount-preview-0")).not.toBeInTheDocument();
+  });
+
+  it("offers a discount editor for prices the plan already has", () => {
+    // Editable, unlike the amount beside it: the discount reaches future subscriptions only, so
+    // changing it cannot contradict anything already sold.
+    render(
+      <Harness>
+        <PlanPriceFields
+          isEditing
+          existingPrices={[price({ automaticDiscountBasisPoints: 800 })]}
+          onUpdatePriceDiscount={async () => {}}
+        />
+      </Harness>,
+    );
+
+    // Named by the price itself, which is also what tells it apart from the draft row's own
+    // field further down the form.
+    expect(screen.getByLabelText(/Automatic discount for €12\.00/)).toHaveValue(8);
+    expect(screen.getByRole("button", { name: /Save discount/ })).toBeInTheDocument();
+  });
+
   it("says nothing about existing prices while creating", () => {
     render(
       <Harness>
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
@@ -28,6 +28,11 @@ import {
   FLAT_FEE,
 } from "../../schemas/subscription-price.schema";
 import { formatPrice } from "../../utilities/subscription-format";
+import {
+  AUTOMATIC_DISCOUNT_COMBINATION_OPTIONS,
+  describeAutomaticDiscount,
+  type AutomaticDiscountCombination,
+} from "../../utilities/subscription-discount";
 import { describeTax, TAX_MODE_OPTIONS } from "../../utilities/subscription-tax";
 import { CardListItem, CardListShell } from "./card-list-shell";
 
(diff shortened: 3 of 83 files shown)


fahad0208:
All four findings addressed in `24ea1ab`. Taking them in the order they matter.

## 1 · Price-restricted promotions surviving a plan change — fixed

Correct diagnosis, including the reason it could not have been fixed at the change site alone: `DiscountTerms` carried no record of what the code had been restricted to, so the change path had nothing to check even if it had thought to look.

- `DiscountTerms` now snapshots `ApplicablePlanCodes` and `ApplicablePriceIds` alongside the amounts, copied at redemption. Read from the snapshot, never the catalogue — a code retired or re-scoped since then is judged by the offer the subscriber accepted.
- New `SubscriptionDiscountApplicability` answers "does this code apply here" once, for both the redemption and the move. Two implementations of that question is how the gap appeared in the first place.
- `SubscriptionPlanChangeService` refuses an ineligible target with `subscription_discount_not_applicable`, as you recommended, rather than silently dropping the promotion: removing it changes what the subscriber pays every period from here on, and an operation they asked for a *price* quote on is the wrong place to do that quietly.
- One deliberate narrowing: the check only applies to a discount that is **still reducing charges**. A spent duration or a passed expiry reduces nothing, so blocking a move over its restrictions would be enforcing an offer that has already ended. Both cases are tested.

Six tests, and the two refusal cases were confirmed to fail against the unfixed guard before it was restored.

## 2 · Invoice-history breakdown — fixed

You were right that one combined figure cannot be turned back into "the price gave 8%, the band gave 5%, the coupon gave nothing". `PaymentDetail` now carries `SubscriptionGrossAmountMinor`, `SubscriptionBuiltInDiscountMinor`, `SubscriptionPromotionalDiscountMinor`, both rates and the combination, filled from the same calculation the charged amount came from; the history response reports them, with the discounted amount **derived** rather than stored so a fourth number cannot disagree with the three it comes from.

The "normally null on a fee invoice" comment was wrong and is gone. What is genuinely absent, now documented as such: a first charge (a hosted checkout composes no invoice) and a plan-change or quantity settlement — the amount there is the difference between two prorated periods and does not decompose into a gross and a reduction. A test pins that those read as `null` rather than as zeroes, since zeroes would claim nothing came off.

## 3 · Unvalidated `applicablePriceIds` — fixed

Both lists are now resolved against the catalogue at authoring time, including the case you did not mention but which fails the same way: a named price sitting on a plan the code is not applicable to, which with both lists narrowing can never match. Rejected with `subscription_discount_applicability_invalid`, naming the offending value — the whole point being that `subscription_discount_not_applicable` at redemption says nothing about the typo behind it.

A price outside the caller's scope reads as *unknown* rather than forbidden, so the refusal cannot be used to confirm that another organization's price exists. Seven tests in a new `DiscountCatalogueServiceTests`, including that an unrestricted discount — the ordinary case — performs no catalogue lookup at all.

## 4 · The rounding explanation — corrected

You are right and I had it exactly backwards. Truncating a *reduction* makes the reduction smaller: 5% of 199 takes off 9 rather than 10, which favours the merchant by up to one minor unit. Truncation stays, because matching the volume bands that are already charging money is the reason for it — but the calculator comment, the README, the integration guide and the client-side preview test now say which way it leans instead of claiming it favours the subscriber.

## One thing I found while in there

A usage invoice was recording `BestDiscount` as its combination. No band takes part in metered overage, so that named a decision the invoice never made, and possibly the wrong one for a price authored as `Additive`. It now records the rate that applied and no combination.

## On the overage question

Agreed that it is a commercial decision rather than a correctness one, and it is worth a product answer before this ships: an 8% discount for annual prepayment does not obviously imply 8% off separately-billed metered overage. The plan's "all charges associated with the selected price" is what I implemented, and it is documented explicitly in both the guide and the README so the decision is visible rather than buried in a calculator. Easy to reverse — it is one call site in `SubscriptionUsageRatingProcessor` — if the answer is no.

## Verification

- Server unit suite **2544 passing**; the 4 `SubscriptionSimulation*` permission-guard failures are pre-existing on `dev` and untouched here.
- Client: typecheck clean, **215/215** in the subscription module, and `npm run build` **passes locally** — noting your point that CI skipped `build-client`, so that check was worth doing by hand.
- The wider client suite has 6 failures in components this branch does not touch or import (payment-provider modals and similar) under a heavily loaded 15-minute run; the subscription module passes in isolation. I have not chased them down and am not claiming they are unrelated on that basis alone — they are simply outside the module this branch changes.

fahad0208:
Both addressed in `c350ba3`.

## 1 · Discount authoring resolved the wrong scope — fixed

Confirmed, and worse than the symptom suggests: `CreateAsync` was the only method in the service passing `null` to the resolver, while `ListAsync` and `ArchiveAsync` both pass the requested organization. It then stored `request.OrganizationId` **unvalidated** — so alongside the false rejections you identified, any caller could persist a discount under another organization's scope. My applicability check turned a latent inconsistency into a visible failure, which is at least how it got found.

- Resolved against `request.OrganizationId` now, so validation runs in the scope the discount is being authored for.
- The persisted scope comes from the resolver's **answer**, not the ask: `null` stays tenant-wide (a scope, not an organization, so it has to survive resolution), and a non-console caller naming somebody else's organization is scoped to its own rather than writing into it.
- Four tests, including the console-mediated case you asked for, which asserts the exact resolver argument:
  ```csharp
  _contextResolver.Verify(r => r.ResolveAsync(It.IsAny<string>(), CustomerOrganizationId, ...), Times.Once);
  _contextResolver.Verify(r => r.ResolveAsync(It.IsAny<string>(), null, ...), Times.Never);
  ```
  Confirmed failing against the old `null` resolution before fixing it.

## 2 · Settlement breakdown — implemented, in the shape you described

You were right on both counts: a single gross/discount tuple cannot describe a settlement, and `null` was not an answer. A settlement now reports the subtraction it actually is, with the fields you listed:

```json
"settlement": {
  "outgoing": { "grossAmountMinor": 1000, "builtInDiscountMinor": 100,
                "promotionalDiscountMinor": 50, "discountedAmountMinor": 850,
                "taxAmountMinor": 85, "periodTotalMinor": 935,
                "proratedValueMinor": 467 },
  "target":   { "grossAmountMinor": 2000, "builtInDiscountMinor": 160,
                "promotionalDiscountMinor": 0, "discountedAmountMinor": 1840,
                "taxAmountMinor": 184, "periodTotalMinor": 2024,
                "proratedValueMinor": 1012 },
  "creditConsumedMinor": 100,
  "netSettlementMinor": 445
}
```

Every figure was already being computed to reach the charge and then thrown away — `SubscriptionProrationCalculator` now returns them instead. Each side is priced as its own period, at its own rate and mode (a change can cross between inclusive and exclusive), and only then prorated.

Three details worth flagging:

- **Snapshotted on the reservation** when the change is quoted, not recomputed at settlement. Recomputing would price a different instant, possibly against an edited catalogue — an explanation of a charge nobody was quoted. It rides the same path as the card and the target terms, which exist for the same reason.
- **`netSettlementMinor` is negative on a downgrade** that banked credit rather than charging, so it is deliberately not "the amount charged". Likewise `creditConsumedMinor` is zero there: reporting the whole balance as consumed would describe money that was not spent.
- **The flat fields and `settlement` are alternatives, never both.** A renewal reports one priced period; a settlement reports two. Tested in both directions, including that the flat fields stay `null` on a settlement rather than describing it badly.

Per-side `discountedAmountMinor` is derived on the way out rather than stored, for the same reason the renewal's is: a fourth number that could contradict the three it comes from is a liability.

## Verification

Server suite **2558 passing**; the 4 `SubscriptionSimulation*` permission-guard failures remain pre-existing on `dev`. Client typecheck clean. The guide and the module README both document the settlement shape and the authoring scope.

fahad0208:
Fixed in `0c0f7ee`, and the fix turned up two more things worth naming.

## The defect

Confirmed exactly as described, including the part about my test. The root cause is that `PlanChangeOrderIdFor` was **dead code** — keyed on a version the reservation-based flow stopped using — so `planchange:` was never written at all. The parser's branch for it looked correct and was unreachable, and every settlement of either kind went out as `quantity:` and came back as `Renewal` with a reservation id sitting where a period key belongs.

- The order id is now kind-aware, with the writer and the reader sharing the same constants so they cannot drift again.
- The **idempotency key is deliberately untouched**. It is what dedupes an attempt, so a reservation taken before this deploy and replayed after it still finds its own charge instead of raising a second. There is a test pinning that the key stayed reservation-only.
- Dead `PlanChangeOrderIdFor` removed rather than left as a second answer to the same question.
- Both legacy spellings are still parsed. Settlements charged before the kinds were distinguished carry the shared form and report `QuantityChange` whichever they were — coarse, but settled financial records do not get rewritten to improve a label, and guessing would be worse than being coarse. Documented in the guide.

## Two things that fell out of it

**1. The segments had to be `pc` and `qty`, not `planchange` and `quantity`.** The existing length test refused `planchange` immediately: a subscription id (36) and a reservation id (32) already spend 68 of the payment module's 80-character order-id limit, so `planchange` lands at 84.

Which exposed the second thing: **the old `quantity` spelling was 82 characters — two over that limit — and had been all along.** The length test checked the renewal id, the usage id and the dead plan-change id, and never the settlement one. `MakePaymentRequestValidator` caps `OrderId` at 80; the subscription gateway writes `PaymentDetail` directly rather than through that validator, which is presumably why nothing rejected it, but it was outside the documented contract. Both kinds now measure 76 and 77, and the test checks both — plus that the two kinds never share a segment, which is the actual invariant behind this bug.

**2. My history test was worse than you said.** It used `settle:`, which nothing produces, *and* asserted the settlement payload without asserting `invoiceType` — so it passed while the classification was wrong, on a shape that never reaches production. All the order-id tests now build their ids from the real helpers, and the new mapping test covers every id this module writes (renewal, usage, plan change, quantity), both legacy spellings, a bare subscription order id and a foreign one, asserting `invoiceType` **and** `periodKey` for each.

## Verification

Server suite **2561 passing**; the 4 `SubscriptionSimulation*` permission-guard failures remain pre-existing on `dev`. The guide now documents every `invoiceType` value, which of them carry `settlement`, and which carry a `periodKey`.

Thank you for the three passes — the plan-change applicability hole and this one were both real money-visible defects that my own tests were structured to miss.

PR #307, opened 2026-08-25
Calendar-aligned monthly billing with a prorated first period

fahad0208:
## What this adds

A monthly price can now choose where its renewal boundary falls:

- **`Anniversary`** — current behaviour. An August 25 signup renews September 25.
- **`CalendarMonth`** — charge a prorated August 25–31 stub, then renew for full months at local midnight on the first.

Selected per price inside the price card in the plan builder, and snapshotted onto every subscription sold on it.

```json
{ "interval": "Month", "intervalCount": 1, "billingAlignment": "CalendarMonth" }
```

`Anniversary` is the default and the enum's zero, so every existing price and subscription deserializes to exactly the behaviour it was sold on.

## How it works

**The recurring cadence needed nothing new.** A calendar-aligned `BillingSchedule` is an ordinary monthly one anchored on the first at local midnight, so `BillingPeriodCalculator` derives every later boundary as it always has. Only the *first* period is special, because it starts mid-month.

**Calendar dates, counted inclusively, in the subscriber's own time zone.** The 25th through the 31st is seven dates over the 31 August has. February uses 28 or 29 as appropriate. The time of day never enters into it — everyone signing up on the 25th pays the same fraction — and a signup on the *local* first gets a whole month that is not reported as prorated.

**The first charge is frozen when the subscription is built**, not derived at settlement. A checkout paid the next morning, resumed later, or picked up by the recovery sweep settles the figure the customer was quoted; a stub priced by the day would otherwise shrink underneath somebody who left the page open overnight.

**The day fraction is applied to the gross once, at the front**, so the volume band, promotional discount and tax all apply to a prorated gross:

1. Gross from price and quantities
2. Prorate by covered calendar days (nearest minor unit, halves away from zero)
3. Band and promotional discount — a *fixed* discount is prorated by the same fraction, a percentage needs no scaling
4. Tax, at the price's own rate and mode
5. Banked credit, last

**Metering is untouched.** The usage schedule keeps the plan's own independent cadence and the allowance stays whole for the stub — an allowance is capacity for a period, not money to be prorated.

## Notable decisions

- **A paid stub consumes one limited-discount period**, but an anniversary first period deliberately still does not. The field's own docs say it should, but making that change here would shorten every existing plan's discount by a period for reasons unrelated to calendar billing.
- **`CalendarMonth` is refused for any other cadence** as `subscription_billing_alignment_invalid`, at authoring time. A quarterly price has no single "first" to renew on that is not also a choice of which month, so there is nothing honest to clamp it to.
- **A snapshot whose alignment and cadence disagree bills by its cadence.** Only reachable by bypassing validation, but it must never bill half-way between the two.
- **Plan changes net two different prorations on purpose.** Unused time on the outgoing plan is credited by *elapsed time* (it is time they paid for); the target stub is priced by *calendar dates* (the same 7/31 a fresh signup would pay).

## New response fields

On subscriptions — populated while checkout is pending, kept afterwards for tracing. `recurringAmountMinor` still means the next *full* month.

```json
{
  "billingAlignment": "CalendarMonth",
  "initialChargeAmountMinor": 3274,
  "initialChargeProrated": true,
  "prorationDays": 7,
  "prorationTotalDays": 31
}
```

## Tests

45 new tests across five server files and three client files, covering the 7/31 case, February in common and leap years, signup on the first, UTC-boundary and daylight-saving dates, the frozen charge surviving a resumed checkout, quantity/tax/discount ordering, fixed-discount proration, the untouched usage allowance, a payment-free trial ending mid-month (including a late-running worker), plan-change netting, the full renewal after a stub, retry-on-the-same-boundary, and anniversary prices being unaffected.

- Server: `0 Error(s)`; full suite shows **no new failures** — the 154 failures are 150 Mongo-dependent `XUnitTest.Integration.*` tests plus 4 pre-existing `SubscriptionSimulationGuard`/permission failures, all present on `dev` before this change.
- Client: `211 passed` across the subscription module, `npm run type-check` clean, `npm run build` clean.

## Docs

`server/Subscription.DomainService/README.md` gains a "Calendar-aligned billing" section covering calendar-day inclusivity and time zones, checkout-creation anchoring, the discount/tax/credit ordering, trials, plan changes, the unchanged invoice boundary, and the validation error.

### DIFF
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-alignment.test.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-alignment.test.tsx
@@ -1,0 +1,108 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, expect, it } from "vitest";
+import { FormProvider, useForm } from "react-hook-form";
+import type { ReactNode } from "react";
+
+import type { PlanPrice } from "../../models/subscription-plan.model";
+import {
+  defaultSubscriptionPlanFormValues,
+  type CreateSubscriptionPlanFormValues,
+} from "../../schemas/subscription-plan.schema";
+import { PlanPriceFields } from "./plan-price-fields";
+
+const Harness = ({ children }: { children: ReactNode }) => {
+  const form = useForm<CreateSubscriptionPlanFormValues>({
+    defaultValues: defaultSubscriptionPlanFormValues,
+  });
+
+  return <FormProvider {...form}>{children}</FormProvider>;
+};
+
+const renderFields = () =>
+  render(
+    <Harness>
+      <PlanPriceFields />
+    </Harness>,
+  );
+
+describe("the billing cycle field", () => {
+  it("is offered for a monthly price, which is the default cadence", () => {
+    renderFields();
+
+    expect(screen.getByLabelText("Billing cycle for price 1")).toBeInTheDocument();
+  });
+
+  it("starts on the anniversary, so an author who ignores it changes nothing", () => {
+    renderFields();
+
+    expect(screen.getByLabelText("Billing cycle for price 1")).toHaveTextContent(
+      "Subscription anniversary",
+    );
+  });
+
+  /**
+   * Hidden rather than disabled. A greyed-out "renew on the 1st" invites the question of how to
+   * enable it, and the answer is to sell a different cadence.
+   */
+  it("disappears once the cadence can no longer carry it", async () => {
+    const user = userEvent.setup();
+    renderFields();
+
+    const howMany = screen.getByRole("spinbutton", { name: /how many/i });
+    await user.clear(howMany);
+    await user.type(howMany, "3");
+
+    expect(screen.queryByLabelText("Billing cycle for price 1")).not.toBeInTheDocument();
+  });
+
+  it("shows the worked example only once the calendar is actually chosen", async () => {
+    const user = userEvent.setup();
+    renderFields();
+
+    expect(screen.queryByTestId("billing-alignment-example-0")).not.toBeInTheDocument();
+
+    await user.click(screen.getByLabelText("Billing cycle for price 1"));
+    await user.click(screen.getByRole("option", { name: /renew on the 1st/i }));
+
+    expect(screen.getByTestId("billing-alignment-example-0")).toHaveTextContent(
+      "A signup on August 25 pays 7/31 of this monthly price, then renews on September 1.",
+    );
+  });
+});
+
+describe("an existing price on the plan", () => {
+  const price = (overrides: Partial<PlanPrice> = {}): PlanPrice => ({
+    priceId: "price-1",
+    currencyCode: "EUR",
+    unitAmountMinor: 1_200,
+    interval: "Month",
+    intervalCount: 1,
+    quantityItemKey: null,
+    ...overrides,
+  });
+
+  it("says when a calendar-aligned one renews", () => {
+    render(
+      <Harness>
+        <PlanPriceFields
+          isEditing
+          existingPrices={[price({ billingAlignment: "CalendarMonth" })]}
+        />
+      </Harness>,
+    );
+
+    expect(screen.getByText(/every month, on the 1st/)).toBeInTheDocument();
+  });
+
+  it("says nothing extra about an anniversary one", () => {
+    render(
+      <Harness>
+        <PlanPriceFields isEditing existingPrices={[price()]} />
+      </Harness>,
+    );
+
+    expect(screen.getByText(/€12\.00 every month/)).toBeInTheDocument();
+    expect(screen.queryByText(/on the 1st/)).not.toBeInTheDocument();
+  });
+});
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
@@ -577,8 +577,10 @@ export const PlanPriceFields = ({
           <p className="mt-2 text-xs text-muted-foreground">
             Retiring stops a price being sold. Anyone already on it keeps their terms and their
             renewals — a subscription bills from what it was sold on, not from this list. Prices
-            keep immutable amount and cadence terms; only tax and automatic-discount metadata can be
-            edited, and only for future subscriptions and future moves onto the price.
+            keep immutable amount, cadence and billing-cycle terms; only tax and
+            automatic-discount metadata can be edited, and only for future subscriptions and future
+            moves onto the price. To move a plan onto a different billing cycle, add a new price and
+            retire this one.
           </p>
         </div>
       )}
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
@@ -18,6 +18,7 @@ import {
   SelectValue,
 } from "@/components/ui-kits/select/select";
 import {
+  BILLING_ALIGNMENT_OPTIONS,
   BILLING_INTERVAL_OPTIONS,
   SUBSCRIPTION_CURRENCY_OPTIONS,
 } from "../../constants/subscription.constants";
@@ -27,6 +28,10 @@ import {
   defaultSubscriptionPriceFormValues,
   FLAT_FEE,
 } from "../../schemas/subscription-price.schema";
+import {
+  CALENDAR_ALIGNMENT_EXAMPLE,
+  isCalendarEligible,
+} from "../../utilities/billing-alignment";
 import { formatPrice } from "../../utilities/subscription-format";
 import {
   AUTOMATIC_DISCOUNT_COMBINATION_OPTIONS,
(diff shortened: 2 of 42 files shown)


fahad0208:
Thanks — all four code-level findings are fixed in 75fea42. The two sequencing items are deferred deliberately; detail below.

## Fixed

**1 (P1) — late trial conversion dropped the stub.** Confirmed, and worse than a shortened period: once the clock crossed the month boundary the days were never billed at all. The conversion now resolves its period from `Trial.EndsAtUtc` rather than the sweep's clock, keyed to the month the trial ended in. A trial ending 20 August discovered on 2 September bills the 12/31 August stub under an August key, advances to 1 September, and leaves the subscription due again so the next pass raises September as a separate charge.

Two tests: one for the stub being billed after the boundary, one asserting the follow-on September charge is full and lands on a *different* idempotency key — the collision would have been the obvious way to get this wrong.

**2 (P1) — activation repriced to decide `DiscountApplied`.** Confirmed. `InitialChargeDiscountApplied` is now frozen alongside the amount at checkout creation, and `StubConsumedDiscountPeriod` is a pure read of two stored fields — activation cannot reprice because it no longer has the machinery to. Test covers a promotion that lapsed between the charge and the webhook.

**4 (P2) — card-free trials recorded a fraction no charge matched.** Confirmed. All initial-charge fields are now left unset at signup for card-free trials and written atomically in the transition that opens the first paid period, via new `SubscriptionTransition` fields. Tests cover both the unset-at-signup case and the 12/31 written at conversion, plus one asserting an ordinary renewal never overwrites what the original checkout froze.

**5 (P2) — plan change on the local first priced by clock.** Confirmed, and the root cause was the one you identified: `BillingDayFraction` conflated "whole" with "absent". It now distinguishes `IsCalendarPriced` from `IsPartial`, `Of` always returns the real day count (`30/30`, not `default`), and every calendar-priced target skips the tick scaling. Theory test at 00:00 and 12:00 on the first, both paying the full month.

## Deferred

**3 (P1) — rebase onto #306.** #306 is still open, so there is nothing on `dev` to rebase onto and rebasing onto its branch would make this PR unmergeable until it lands. Agreed this must happen before the combined financial behaviour is verified, and agreed on the ordering:

```
full gross → calendar-day proration → automatic + quantity built-in → promotional → tax → credit
```

Worth noting the current design should compose cleanly: the fraction is applied to the gross once at the front of `DiscountedAmountMinor`, so a `BuiltInDiscountCalculator` inserted after that point automatically operates on a prorated gross. The risk you flagged — automatic discounts applied before proration — would require moving the scaling, not preserving it. I will do the rebase and add the `BestDiscount` / `Additive` / automatic-only / automatic-plus-promo stub tests as soon as #306 is on `dev`.

**6 (P3) — alignment editor for existing prices.** Taking the "explicitly narrow the scope" option for now, since #306 rewrites the price-card UI and price models and an editor built here would largely be reworked by the rebase. The gap is no longer silent: the price card says alignment is immutable alongside amount and cadence and points at add-new-price-and-retire, and the README says the same and why. Happy to build the CAS-protected future-snapshots-only endpoint on the rebased branch if you would rather it ship with this feature than after it.

## Verification

- Server: `0 Error(s)`; full suite **154 failures, unchanged from before these fixes** — 150 Mongo-dependent `XUnitTest.Integration.*` plus the 4 pre-existing `SubscriptionSimulationGuard`/permission failures that are also red on `dev`.
- 13 new tests (58 → 71 across the calendar files, plus 3 on `SubscriptionActivationProcessorTests`).
- Client: 211 passed, `type-check` clean, `build` clean.

fahad0208:
All three findings addressed. The branch is rebased onto `dev` with #306 merged and now reports `MERGEABLE`.

## 1 (P1) — declined trial stub abandoned on retry

Confirmed, and your reading of the mechanism was exact: `ApplyFailureAsync` moves a declined conversion to `PastDue`, after which the status-keyed check returned false and the retry billed whatever month the clock had reached.

Recognition no longer asks about status at all. What actually ends a conversion is its first paid period being recorded, so the check is now `InitialChargeAmountMinor is null` plus the trial terms — which is true through any number of dunning attempts and false the moment the stub is paid, whatever status it lands in.

You were also right about the test. `A_retry_after_a_decline...` ran two successful `Active` renewals and proved nothing about dunning. The gateway stub can now be told to decline, and it exercises a real decline into `PastDue` and back. Added alongside it:

- `A_declined_conversion_still_retries_the_stub_it_owes` — your exact scenario, asserting 12/31 on the retry, the August period identity, the first-charge fields recorded on success, and September left due separately.
- `A_paid_conversion_is_not_retried_as_one` — the other side of the predicate.

## 2 (P1) — late processing changed discount eligibility

Confirmed. The period moved to the trial end but pricing still passed `now`, so `DiscountStillActive` was evaluated against the sweep's clock — the same contractual period costing two different amounts depending on worker latency.

Split as you suggested: `pricingInstantUtc` is `trialEndUtc` for a conversion and `now` for everything else, while the operational instant stays `now` for audit, dunning and scheduling. Test has a promotion live on 20 August and lapsed by the 2 September sweep, asserting the discounted 2756 rather than the undiscounted 3445.

## 3 (P1) — composition with #306

Rebased onto `5c7c635`. Conflicts were where you predicted; the substantive one was `SubscriptionAmountCalculator`.

The composition point held — `BuiltInDiscountCalculator.Resolve` takes gross as a parameter, so prorating in front of it puts the automatic rate on an already-prorated gross with no reordering.

One thing your list did not call out, and it is the real hazard in this merge: **`BuiltInDiscountCalculator` uses a volume band's resolved *money* verbatim when the band wins.** That is deliberate on #306's part — it is what keeps pre-existing bands priced to the same minor unit — but the figure is a whole month's. Passing a prorated gross alongside an unprorated band would have taken a full month's volume discount off a fraction of a month's charge; a 15% band on a 7/31 stub would have subtracted 16020 from 24116 instead of 3617. The band's *rate* is now re-expressed against the prorated gross for a partial period and passed through untouched for a whole one, so anniversary pricing is bit-for-bit unchanged. `A_winning_volume_band_is_re_expressed_against_the_prorated_gross` locks it.

Final ordering, documented in the README:

```
full gross → calendar-day proration → automatic + quantity built-in → promotional → tax → credit
```

New `CalendarAlignedAutomaticDiscountTests` covers every case you listed — automatic-only, automatic + promotional, `BestDiscount` in both directions, `Additive`, `Stack`, fixed-code proration alongside an unprorated automatic rate, tax after both, and a plan change onto a calendar-aligned automatically discounted price — plus a promotional code that loses to the automatic rate and is correctly not consumed.

## Verification (post-rebase, so this is the combined code)

- Server: `0 Error(s)`. Full suite **2806 tests, 154 failures — unchanged** from before the rebase and from `dev`: 150 Mongo-dependent `XUnitTest.Integration.*` plus the 4 pre-existing `SubscriptionSimulationGuard`/permission failures.
- Calendar suite: **84 passed, 0 failed**.
- Client: 236 passed across the subscription module, `type-check` clean, `build` clean.

Alignment immutability stays the accepted limitation, now worded to sit correctly alongside #306's editable automatic-discount metadata.

fahad0208:
Calendar alignment now extends to yearly prices, in `5c615cb`.

## What it does

A `Year` × 1 price can be `CalendarMonth`-aligned. Tier 2 signing up 25 August:

| | Calculation | Amount |
|---|---|---|
| 25–31 August | `95000 × 7/31` = 21452, less 8% | **CHF 197.36** |
| 1 September | `95000 × 12` = 1140000, less 8% | **CHF 10,488.00** |
| 1 September next year | the same again | **CHF 10,488.00** |

## The two decisions worth reviewing

**A year anchors on the first of the *next* month, not the current one.** This is the substance of
the change rather than a detail. A year anchored on the month it started in would end on the
1 August after a 25 August signup — eleven months for a year's money — and no later boundary could
correct it, because every one derives from the anchor. Monthly keeps anchoring on the month it is
already inside, so its opening period is the remainder of that month.

**The stub is priced from a linked monthly price, not from a daily rate.** A month's stub is a
fraction of the very price being charged; a year's cannot be, because a week of an annual amount is
not a quantity anybody can charge. `calendarStubBasePriceId` is required for calendar-aligned
`Year` × 1 and refused everywhere else, and the referenced price is validated against plan, cadence,
currency, quantity item and tax configuration — each of those, left to differ, produces two figures
a subscriber cannot reconcile and only discovers on an invoice.

Naming it makes `unitAmountMinor` server-derived at twelve times the monthly amount. A conflicting
submitted amount is refused (`subscription_calendar_stub_base_price_amount_conflict`) rather than
overwritten, and the plan builder shows the annual figure read-only rather than offering an edit
that would be discarded.

## Two things the tests caught that I would flag

**The plan-change path bypassed the stub basis.** `SubscriptionProrationCalculator` calls
`DiscountedAmountMinor` directly rather than through the subscription, so a monthly→yearly change
was prorating the *annual* amount by days — CHF 2,153.75 where the correct settlement was a
CHF 17.15 credit. Fixed, and covered.

**A yearly snapshot with no monthly basis has to fail closed.** With the basis missing, the fraction
still applied and the annual amount was prorated by days: a week billed at roughly a twelfth of its
worth. `IsCalendarAligned(PriceSnapshot)` now requires the basis for cadences that need one, so such
a price falls back to an ordinary anniversary year. Only reachable through legacy or corrupt data,
but it is the one direction worth failing closed in.

## Snapshot immutability

The monthly amount and price id are copied onto the subscription's `PriceSnapshot`. Nothing reads
the monthly price after authoring — not checkout, not renewal, not recovery. The stub is charged at
checkout and the annual period a month later, so a live read would let somebody editing the monthly
price in between change what an annual subscriber already agreed to. Covered by a test that reprices
the catalogue afterwards and asserts the quoted figure is unchanged.

## Verification

- Server: `0 Error(s)`. Full suite **2842 tests, 154 failures — unchanged** from before this change:
  150 Mongo-dependent `XUnitTest.Integration.*` plus the 4 pre-existing `SubscriptionSimulationGuard`
  failures, all also red on `dev`.
- Calendar suite: **102 passed, 0 failed**, including trial conversion, declined-conversion dunning
  and monthly→yearly plan change on the yearly path.
- Client: 246 passed across the subscription module, `type-check` clean, `build` clean.

## Not covered

Separate stub and annual invoice records are not implemented — invoice history still begins at the
first renewal, as documented for the monthly feature. Say the word if you want that in this PR
rather than after it.

PR #313, opened 2026-08-25
Calendar-aligned yearly billing: state and charging

fahad0208:
Implements the **billing-state half** of the Configurable Calendar-Aligned Yearly Billing plan. Invoice snapshots are the other half and are tracked in #315, which blocks the plan being complete.

Follows #307 (monthly calendar alignment) and sits on top of #306 and #310.

## Three ways to sell a year

Configured per price, so one plan can offer more than one and an organization-scoped catalogue can differ. Tier 2 — CHF 950 a month, CHF 11,400 a year authored independently, 8% off for paying annually — signing up 25 August:

| Mode | At checkout | On 1 September | Renews | Cancelling during the stub |
|---|---|---|---|---|
| `Anniversary` | CHF 10,488.00 | — | 25 August | ordinary rules |
| `CalendarMonth` + `AtBoundary` | CHF 197.36 | CHF 10,488.00 charged | 1 September | access ends with the stub; the year is never charged |
| `CalendarMonth` + `AtCheckout` | CHF 10,685.36 | year opens, nothing charged | 1 September | nothing refunded; access runs to the end of the year |

Both calendar modes come to the same money. They differ in when it is taken and in what a cancellation leaves the subscriber holding — which makes the choice a refund policy as much as a collection date, so the plan builder states both consequences rather than only the timing.

## Contract

- `billingAlignment`: `Anniversary` | `CalendarMonth`, now valid for `Month × 1` **and** `Year × 1`.
- `calendarStubBasePriceId`: the monthly price a yearly stub is a fraction of. Required for calendar-aligned yearly, refused elsewhere.
- `calendarAnnualChargeTiming`: `AtBoundary` | `AtCheckout`. Same cadence rule.

**The annual `unitAmountMinor` is authored, not derived.** What a year costs is a commercial decision and an annual plan is usually not twelve monthly ones; the linked monthly price prices the opening stub and nothing else.

The linked price is validated at authoring time — same plan, active, `Month × 1`, same currency, same quantity item, same tax rate and mode — and the price-card picker applies every one of those rules so it cannot offer a basis the API rejects.

## Discounts

The yearly price's automatic discount and volume band apply to the stub and the year. A **promotional code applies to the year alone** and is consumed once: spending a month of a three-month promotion on a seven-day stub would exchange a month of the customer's discount for a week of it.

## State

`PendingAnnualPeriod` carries the year between signup and its boundary — dates, full breakdown, whether a promotion reduced it, and two separate flags:

- `CollectedWithCheckout` — what the price says to bill for, written at signup.
- `IsPrepaid` — whether the money arrived, written only by the activation that records the opening payment.

The boundary skips the gateway on the second, never the first. Deriving it from configuration would let an unpaid checkout open a year nobody paid for, and would have `/current` report an incomplete checkout as settled.

Every figure is frozen at checkout and none is recalculated at the boundary a month later. Opening the year and discarding the pending record are one transition — a boundary that did the first and not the second would charge for the same year twice.

**Plan and quantity changes are refused while a year is pending** (`409 subscription_initial_annual_period_pending`).

## Trials

A card-free trial's year is priced at conversion, not at signup: a 25 August signup on a trial ending 20 September starts its year on 1 October. The schedule is anchored on the trial's end for the same reason — every boundary derives from the anchor, and one anchored a month early stays a month early forever. On an `AtCheckout` price the conversion is the first payment there has ever been, so it takes the stub and the year together.

## Verification

- Server: `0 Error(s)`. Full suite **2871 tests, 154 failures — unchanged from `dev`**: 150 Mongo-dependent `XUnitTest.Integration.*` plus 4 pre-existing `SubscriptionSimulationGuard`/permission failures.
- Client: 253 passed, `type-check`, `lint` (no new findings in touched files) and `build` clean.
- `server/Subscription.DomainService/README.md` and the maintained API guide both document the three-mode contract.

## Out of scope

Invoice snapshots, the two-line `AtCheckout` invoice, and history/PDF wiring — **#315**, with acceptance criteria. Invoice history remains payment-derived and begins at the first renewal.

### DIFF
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
@@ -34,6 +34,7 @@ import {
   CALENDAR_ALIGNMENT_EXAMPLE,
   CALENDAR_YEARLY_ALIGNMENT_EXAMPLE,
   isCalendarEligible,
+  isCompatibleStubBasis,
   needsStubBasePrice,
   requiresStubBasePrice,
 } from "../../utilities/billing-alignment";
@@ -430,9 +431,9 @@ const PriceDiscountFields = ({ index }: { index: number }) => {
 /**
  * What the price costs, entered or derived.
  *
- * A calendar-aligned yearly price linked to a monthly one has no amount of its own: the server
- * derives it as twelve times the monthly figure, so the field is shown read-only rather than
- * offering an edit that would be discarded. Everything else is an ordinary amount input.
+ * Always the author's to set. A calendar-aligned yearly price is linked to a monthly one, but that
+ * link prices the opening period only — what a year costs stays a commercial decision, and an
+ * annual plan is usually not twelve monthly ones.
  */
 const PriceAmountField = ({ index }: { index: number }) => {
   const { control } = useFormContext<CreateSubscriptionPlanFormValues>();
@@ -575,8 +576,17 @@ const PriceStubBasisField = ({
 }) => {
   const { control } = useFormContext<CreateSubscriptionPlanFormValues>();
   const price = useWatch({ control, name: `prices.${index}` });
-  const eligible = monthlyPrices.filter(
-    (candidate) => candidate.currencyCode === price?.currencyCode,
+  // Every rule the server applies, so the picker cannot offer a choice the API refuses.
+  const eligible = monthlyPrices.filter((candidate) =>
+    isCompatibleStubBasis(candidate, {
+      currencyCode: price?.currencyCode ?? "",
+      quantityItemKey: price?.quantityItemKey,
+      taxPercent:
+        price?.taxPercent === undefined || price?.taxPercent === null
+          ? undefined
+          : Number(price.taxPercent),
+      taxMode: price?.taxMode,
+    }),
   );
 
   const basis = eligible.find(
@@ -589,8 +599,9 @@ const PriceStubBasisField = ({
 
       {eligible.length === 0 ? (
         <p className="text-xs text-muted-foreground" data-testid={`stub-basis-empty-${index}`}>
-          This plan has no monthly {price?.currencyCode} price yet. Add and save one first — a
-          yearly price on the calendar is charged from it, and derives its own amount from it.
+          This plan has no compatible monthly {price?.currencyCode} price yet. The opening period
+          is charged from one, so it has to match this price&apos;s currency, quantity item and tax
+          before it can be chosen. Add and save one first.
         </p>
       ) : (
         <>
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
@@ -429,11 +429,12 @@ const PriceDiscountFields = ({ index }: { index: number }) => {
 };
 
 /**
- * What the price costs, entered or derived.
+ * What the price costs.
  *
- * Always the author's to set. A calendar-aligned yearly price is linked to a monthly one, but that
- * link prices the opening period only — what a year costs stays a commercial decision, and an
- * annual plan is usually not twelve monthly ones.
+ * Always the author's to set, on every cadence. A calendar-aligned yearly price is linked to a
+ * monthly one, but that link prices the opening period only — what a year costs stays a commercial
+ * decision, and an annual plan is usually not twelve monthly ones. Only the wording of the hint
+ * changes, to say which of the two figures this one is.
  */
 const PriceAmountField = ({ index }: { index: number }) => {
   const { control } = useFormContext<CreateSubscriptionPlanFormValues>();
@@ -444,7 +445,6 @@ const PriceAmountField = ({ index }: { index: number }) => {
     intervalCount: Number(price?.intervalCount),
     billingAlignment: price?.billingAlignment,
   });
-  const derived = yearlyCalendar;
 
   return (
     <FormField
@@ -464,7 +464,7 @@ const PriceAmountField = ({ index }: { index: number }) => {
             />
           </FormControl>
           <FormDescription className="text-xs">
-            {derived
+            {yearlyCalendar
               ? "What a year costs, in major units. The monthly price below prices only the opening period."
               : "In major units — 89.00, not 8900."}
           </FormDescription>
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-price-fields.tsx
@@ -21,6 +21,7 @@ import {
   BILLING_ALIGNMENT_OPTIONS,
   BILLING_INTERVAL_OPTIONS,
   SUBSCRIPTION_CURRENCY_OPTIONS,
+  YEARLY_BILLING_ALIGNMENT_OPTIONS,
 } from "../../constants/subscription.constants";
 import type { PlanPrice } from "../../models/subscription-plan.model";
 import type { CreateSubscriptionPlanFormValues } from "../../schemas/subscription-plan.schema";
@@ -30,9 +31,13 @@ import {
 } from "../../schemas/subscription-price.schema";
 import {
   CALENDAR_ALIGNMENT_EXAMPLE,
+  CALENDAR_YEARLY_ALIGNMENT_EXAMPLE,
   isCalendarEligible,
+  MONTHS_IN_A_YEAR,
+  needsStubBasePrice,
+  requiresStubBasePrice,
 } from "../../utilities/billing-alignment";
-import { formatPrice } from "../../utilities/subscription-format";
+import { formatMoney, formatPrice } from "../../utilities/subscription-format";
 import {
   AUTOMATIC_DISCOUNT_COMBINATION_OPTIONS,
   describeAutomaticDiscount,
@@ -423,6 +428,54 @@ const PriceDiscountFields = ({ index }: { index: number }) => {
 };
 
 /**
+ * What the price costs, entered or derived.
+ *
+ * A calendar-aligned yearly price linked to a monthly one has no amount of its own: the server
+ * derives it as twelve times the monthly figure, so the field is shown read-only rather than
+ * offering an edit that would be discarded. Everything else is an ordinary amount input.
+ */
+const PriceAmountField = ({ index }: { index: number }) => {
+  const { control } = useFormContext<CreateSubscriptionPlanFormValues>();
+  const price = useWatch({ control, name: `prices.${index}` });
+
+  const derived = requiresStubBasePrice({
+    interval: Number(price?.interval),
+    intervalCount: Number(price?.intervalCount),
+    billingAlignment: price?.billingAlignment,
+  });
+
+  return (
+    <FormField
+      control={control}
+      name={`prices.${index}.amount`}
+      render={({ field: inputField }) => (
+        <FormItem>
+          <FormLabel className="text-xs">Amount</FormLabel>
+          <FormControl>
+            <Input
+              {...inputField}
+              type="number"
+              min={0}
+              step="0.01"
+              placeholder="89.00"
+              readOnly={derived}
+              aria-label={`Amount for price ${index + 1}`}
+              className={derived ? "bg-muted text-muted-foreground" : undefined}
+            />
+          </FormControl>
+          <FormDescription className="text-xs">
+            {derived
+              ? "Derived from the monthly price below — twelve times its amount."
+              : "In major units — 89.00, not 8900."}
+          </FormDescription>
+          <FormMessage />
+        </FormItem>
+      )}
+    />
+  );
+};
+
+/**
  * When this price renews, for the one cadence that gets a choice.
  *
  * Hidden rather than disabled for every other cadence. A greyed-out "renew on the 1st" invites the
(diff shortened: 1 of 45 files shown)


fahad0208:
Review against the Calendar-Aligned Yearly Billing with a Monthly Stub plan:

1. **[P1] Missing invoices for the two-charge lifecycle.** The PR explicitly leaves separate stub and annual invoice records out. The feature creates a paid stub at checkout and a full-year charge on the next first, while the agreed invoice requirement is one invoice per transaction with identity, period, subtotal, discounts, tax, trial and total. The opening stub remains absent from invoice history. Include first-charge invoicing here, or track it as a blocking linked follow-up and do not call the plan complete.

2. **[P1] Annual derivation can overflow into a negative price.** `PlanCatalogueService.cs:266` performs `stubBasePrice.UnitAmountMinor * 12` as unchecked `long` arithmetic after validating only the submitted request amount. A monthly value above `long.MaxValue / 12` wraps and the derived result is persisted without another currency/positivity check. Use checked arithmetic, return a named validation failure, validate the derived amount, and add a boundary test.

3. **[P2] The UI offers monthly bases the server must reject.** `plan-price-fields.tsx:579` filters only by currency, while the server also requires matching quantity item, tax rate and tax mode. With multiple CHF monthly prices, invalid choices appear and fail only after the edit partially saves. Filter by all compatibility rules, or disable incompatible choices with a reason, and test quantity/tax mismatches.

4. **[P2] The integration guide documents the opposite contract.** `subscription/v1/index.html:1600` still says `CalendarMonth` is valid only for a one-month cadence and has no `calendarStubBasePriceId` example. Update the request/response contract, validation errors, derived annual amount and Aug-25 to Sep-1 lifecycle.

The schedule, snapshot, trial/dunning and proration implementation otherwise looks coherent, with strong lifecycle coverage.

fahad0208:
Pushed `ef59b63`. The yearly feature is now configurable per price, per the revised spec, and I have retitled this PR to match.

## Three modes, one signup

Tier 2 — CHF 950 a month, CHF 11,400 a year, 8% off for paying annually — signing up 25 August:

| Mode | At checkout | On 1 September | Renews |
|---|---|---|---|
| `Anniversary` | CHF 10,488.00 | — | 25 August |
| `CalendarMonth` + `AtBoundary` | CHF 197.36 | CHF 10,488.00 charged | 1 September |
| `CalendarMonth` + `AtCheckout` | CHF 10,685.36 | year opens, nothing charged | 1 September |

Both calendar modes come to the same money. They differ in when it is taken and in what cancelling during the stub leaves the subscriber holding — which makes the choice a refund policy as much as a collection date, so the builder states both consequences rather than only the timing.

## Reversal from the previous commit, on purpose

The annual amount is **authored again** rather than derived as twelve times the monthly one. Your spec is right and my earlier reasoning was not: what a year costs is a commercial decision, an annual plan is usually not twelve monthly ones, and deriving it took that decision away from whoever is selling it. The linked monthly price now prices the opening stub and nothing else. The conflict error and the read-only amount field are gone.

## Two decisions worth review

**A promotional code applies to the year alone.** Spending a month of a customer's three-month promotion on a seven-day stub would exchange a month of their discount for a week of it. The price's own automatic discount and volume band still apply to both, because those are properties of the price rather than something the customer spends — so `FirstPeriodCharge` gained an `includePromotionalDiscount` flag rather than the stub being priced through a separate path.

**Prepaid means the boundary takes nothing.** `PendingAnnualPeriod.IsPrepaid` is what stops the same year being billed twice, and it is the single most dangerous flag in the change — so `A_prepaid_year_opens_without_taking_anything` asserts the gateway is not called at all, not merely that the amount is zero.

## The year in between

`PendingAnnualPeriod` carries the year's dates, full breakdown, whether a promotion reduced it, and whether it is paid. Every figure is frozen at checkout and none is recalculated at the boundary a month later. Opening the year and discarding the pending record are one transition — a boundary that did the first and not the second would find the year again on the next sweep and charge for it twice.

A declined boundary charge leaves the year pending and enters ordinary dunning, so the retry owes exactly the frozen amount.

**Plan and quantity changes are refused while a year is pending** (`subscription_initial_annual_period_pending`). Repricing then would have to unpick a settled annual charge or silently discard one about to be collected, and neither is something a caller can be told about afterwards.

**Cancellation** during a prepaid stub runs entitlement to the end of the year they bought; during an unpaid stub it ends with the stub and the year is never charged.

## Verification

- Server: `0 Error(s)`. Full suite **2863 tests, 154 failures — unchanged** from `dev`: 150 Mongo-dependent `XUnitTest.Integration.*` plus the 4 pre-existing `SubscriptionSimulationGuard`/permission failures.
- 16 new tests across the three modes and the boundary, including prepaid double-charge, frozen-amount-at-boundary, decline-and-retry, and the second year renewing normally.
- Client: 250 passed, `type-check`, `lint` (no new findings in touched files) and `build` all clean.

## Not included, by agreement

The invoice-snapshot subsystem. Separate stub and annual invoices already fall out of `AtBoundary` making two payments, and `AtCheckout` produces one invoice for both. What is **not** built: line-level breakdown of that combined invoice, persisted immutable invoice snapshots, and their wiring into the history and PDF endpoints. Invoice history is still derived from settled payments and begins at the first renewal. That follows in its own PR.

fahad0208:
Review of ef59b63 against the Configurable Calendar-Aligned Yearly Billing plan:

1. **[P1] Card-free trials freeze and later open the wrong pending year.** `SubscriptionCreationService.ApplyPeriods` builds `PendingAnnualPeriod` at signup (`SubscriptionCreationService.cs:458`) even though a card-free trial's first paid period must be derived from `Trial.EndsAtUtc`. For a signup on 25 August with a trial ending 20 September, the stored year still starts 1 September. At conversion, `DueAnnualPeriod` sees that stale record as due and can skip the 20–30 September stub/open the September year. The trial tests construct `SubscriptionDetail` manually without the pending record, so they do not exercise the state produced by `CreateAsync`. Do not create the pending year for a card-free trial at signup; create/freeze it atomically when conversion resolves the stub, and add an end-to-end creation→conversion test crossing a month boundary for both charge timings.

2. **[P1] Immediate cancellation can discard a prepaid year without refund or access.** `CancellationService.cs:107` honors `immediately=true` for an active AtCheckout subscription. `EndNowAsync` ends access now and clears the pending prepaid year, although the combined checkout already collected it. The agreed rule is no refund and access through the prepaid annual end. Reject immediate cancellation while a prepaid year is pending, or coerce it to period-end cancellation; test both the default and `immediately=true` calls.

3. **[P1] `IsPrepaid` means configured timing, not successful payment.** `BuildPendingAnnualPeriod` sets it true before checkout is paid, and `PendingAnnualPeriodResponse` documents it as 'was collected'. Therefore `/current` reports an incomplete unpaid checkout as prepaid. The same flag is also the double-charge guard. Store charge timing separately from settlement state, and mark the pending year prepaid in the activation transition that records `InitialPaymentDetailId`; the boundary must skip the gateway only when that paid transition landed.

4. **[P2] Cancellation returns a response different from the state written.** The transition extends `CurrentPeriodEndUtc` to the prepaid year's end and clears `PendingAnnualPeriod`, but `Reflect` (`CancellationService.cs:216`) applies neither change. The 200 response still shows the August stub ending and a pending year even though Mongo now says access lasts through next year and no year is pending. Reflect those fields or re-read after the CAS, and add a response assertion.

5. **[P2] The UI still offers monthly bases that the API rejects.** `plan-price-fields.tsx` filters candidates only by currency. The server also requires the same quantity item, tax rate, and tax mode. Filter/disable using all compatibility rules and test flat-vs-seat and tax mismatches. Several comments in this component and `PlanCatalogueService` also still say the annual amount is derived/twelve times monthly, which is now false.

6. **[P2] The maintained guide and PR body are stale.** `subscription/v1/index.html:1600` still says CalendarMonth supports only monthly cadence and documents neither `calendarAnnualChargeTiming` nor `calendarStubBasePriceId`. The PR body still describes monthly×12 derivation. Update both to the actual three-mode contract and lifecycle.

7. **Plan completeness: invoice work is explicitly absent.** The agreed plan requires one invoice per transaction and two lines for AtCheckout. This PR still has payment-derived history beginning at renewal, so it does not fully implement the plan. If intentionally split, link a blocking follow-up with acceptance criteria and describe #313 as billing-state work, not the complete plan.

The Anniversary, AtBoundary and AtCheckout happy-path arithmetic/scheduling are otherwise represented cleanly, and the boundary test that asserts no gateway call for a settled prepaid year is the right invariant.

fahad0208:
All six findings fixed in `624e3aa`; the seventh is now tracked as **#315** with acceptance criteria, and this PR is retitled and rewritten to describe the billing-state half rather than the whole plan.

## 1 — Card-free trials froze the wrong year

Confirmed, and worse than a wrong stored date: your 25 August / 20 September example would have had conversion see the stale 1 September record as due, open the year, and never bill the 20–30 September stub at all.

The year is now priced at conversion, atomically with the stub it follows. But that alone was not enough — **the schedule was also anchored on signup**, so even a correctly-timed conversion produced a year starting 1 September. The trial's end is known at signup, so the anchor now comes from it. Every boundary derives from the anchor, and one anchored a month early stays a month early forever.

One case your finding implies that I had to decide: a card-free trial on an `AtCheckout` price has no checkout to collect the year with. Its conversion is the first payment there has ever been, so it takes the stub and the year together and settles both.

Covered end-to-end through `CreateAsync` → conversion, crossing a month boundary, for both timings — as you asked, driven from what creation actually produces rather than a hand-built subscription.

## 2 — Immediate cancellation could discard a paid year

Confirmed. Taking the money and the entitlement together is the one outcome no reading of "cancel now" justifies.

I took the coercion option rather than the rejection: while a **settled** year is pending, an immediate request is honoured as far as it can be — cancelled, no renewal, access to the end of the term they bought. Rejecting would leave a caller unable to cancel at all. An unpaid year still takes the ordinary path and is simply dropped. Tested as a theory over both `immediately` values.

## 3 — `IsPrepaid` meant configuration, not payment

Confirmed, and this was the most serious of the three: the same flag was both the API's claim about what the subscriber owes and the guard against double-charging a year.

Split as you specified. `CollectedWithCheckout` is what the price says to bill for, written at signup. `IsPrepaid` is whether the money arrived, written only by the activation transition that records `InitialPaymentDetailId`. The boundary skips the gateway on the second, never the first — so a checkout nobody pays cannot open a year nobody paid for.

## 4 — Cancellation response disagreed with the write

Confirmed. `Reflect` now applies both the extended period end and the cleared pending year, and the test asserts on the response a caller actually receives — built with the real `SubscriptionResponseMapper` rather than a mock the test could agree with itself about.

## 5 — UI offered bases the API rejects

Confirmed. `isCompatibleStubBasis` now mirrors the server's rule exactly — currency, quantity item, tax rate and mode — with the flat-fee sentinel handled and an absent tax mode read as exclusive. Tested for quantity-item and tax mismatches, and the empty-state copy says what compatibility means. The stale monthly × 12 comments are gone from the component and from `PlanCatalogueService`.

## 6 — Guide and PR body stale

The guide gains a "Calendar-aligned yearly prices" section: both new fields, a worked example, the two-timing table with cancellation consequences, every validation error code, and the pending-year change block. The PR body no longer describes the derivation and no longer claims to implement the whole plan.

## 7 — Invoice work

Split as agreed before this PR was written, but you are right that it was described too lightly. **#315** now carries it with eight acceptance criteria, states why it is architectural rather than incremental — invoice history is projected from `PaymentDetail`, and the checkout path produces no Stripe invoice, so first-period charges cannot appear at all today — and is marked as blocking the plan's completion.

## Verification

- Server: **2871 tests, 154 failures — unchanged** from `dev` (150 Mongo integration + 4 pre-existing simulation-permission).
- Client: 253 passed; `type-check`, `lint` and `build` clean.

fahad0208:
Re-review of `624e3aa` against the previous findings:

The six code/documentation findings are fixed in the intended shape. In particular, creation no longer freezes a card-free trial's year, the schedule anchors on the known trial end, activation is the only writer that turns configuration into paid state, prepaid immediate cancellation is safely coerced, the returned cancellation response mirrors the write, the UI compatibility predicate matches server validation, and the guide/PR body now describe the three actual modes. #315 also captures the deferred invoice requirement with appropriate acceptance criteria and explicitly blocks the full plan.

One new money-state defect remains:

**[P1] An AtCheckout promotion is consumed twice.** For an ordinary AtCheckout signup, `InitialChargeDiscountApplied` is true when the annual promotional discount reduced the combined opening charge. Activation therefore increments `DiscountPeriodsApplied` once through `StubConsumedDiscountPeriod`. At the first-of-month boundary, `SubscriptionRenewalService` creates the zero-amount prepaid `PeriodCharge` with `DiscountApplied = annual.DiscountApplied`; `ApplySuccessAsync` increments `DiscountPeriodsApplied` again even though the gateway is correctly not called. The card-free-trial AtCheckout path has the same shape: conversion persists the discounted prepaid year and consumes once, then its no-charge boundary consumes again. A one-period promotion therefore counts as two periods and expires one renewal early.

For a prepaid boundary transition, the discount was already consumed by the payment that prepaid the year, so the zero-amount transition must carry `DiscountApplied = false` (while retaining the frozen breakdown for presentation), or `ApplySuccessAsync` must explicitly suppress consumption when opening an already-prepaid period. Add assertions on `DiscountPeriodsApplied` across activation/conversion and the following no-charge boundary for both ordinary and card-free-trial AtCheckout subscriptions. Keep the AtBoundary assertion showing it consumes exactly once when the annual charge actually succeeds.

Minor cleanup: the doc comment above `PriceStubBasisField` still says the annual amount is derived/read-only even though the field is now authored. It is non-functional but directly contradicts the corrected contract.

With the double-consumption fixed, I have no remaining billing-state objection to #313; invoice completeness remains intentionally blocked by #315.

fahad0208:
Both fixed in `bce5754`.

## The double consumption

Confirmed exactly as described. `SubscriptionRenewalService` built the boundary charge from the frozen record and passed `annual.DiscountApplied` through unconditionally, and `ApplySuccessAsync` increments `DiscountPeriodsApplied` on that flag — so the prepaid transition counted a period while charging nothing.

The rule is now stated where it is applied: **a promotion is spent by the charge that reduced money.** A prepaid year reduced it once already, at the checkout or the trial conversion that collected it, so the transition that merely opens it reports no discount applied. A year still owed is the opposite case — its boundary *is* the charge that spends it, and that is where it counts.

```csharp
DiscountApplied: !annual.IsPrepaid && annual.DiscountApplied,
```

Worth noting the defect was not confined to ordinary checkout: the **card-free trial conversion on an at-checkout price** had the same shape. That conversion collects the stub and the year together and reports the discount applied, then the boundary opened the prepaid year and counted it again. One fix covers both, because both go through the same prepaid-opening branch.

## Tests

`CalendarAnnualDiscountConsumptionTests`, covering the three cases you asked for:

- **Prepaid boundary** — the defect. Starts from a subscription whose checkout already spent the period, and asserts the count does not move.
- **At-boundary control** — must still increment, since that boundary is the charge.
- **End to end across both transitions** — drives the real activation processor and then the boundary, asserting a one-period promotion is gone after exactly one bill rather than none or two.

The trial-conversion route is covered by the existing `A_prepaid_trial_conversion_collects_the_stub_and_the_year_together` plus the shared branch these exercise; I did not add a fourth near-duplicate for it.

## Stale comment

Gone, along with the `derived` alias it named — that variable had outlived its meaning and only ever selected between two hint strings. It is now `yearlyCalendar`, which is what it actually tests.

## Verification

- Server: `0 Error(s)`. Full suite **2874 tests, 154 failures — unchanged from `dev`**: 150 Mongo-dependent `XUnitTest.Integration.*` plus the 4 pre-existing `SubscriptionSimulationGuard`/permission failures.
- Client: 253 passed; `type-check` and `build` clean.

Invoice completeness remains with **#315** as agreed.

fahad0208:
Re-review of `bce5754`:

The latest finding is fixed correctly. `DiscountApplied: !annual.IsPrepaid && annual.DiscountApplied` ties consumption to the charge that moved money: a prepaid boundary consumes zero, while an AtBoundary annual charge consumes once. The stale UI naming/comment is also removed.

The new tests prove those isolated transitions, but the activation-to-boundary test does not actually carry persisted state from activation into renewal: it captures the activation transition, then constructs a fresh `InStub(prepaid: true)` object. That is acceptable for the specific boolean rule, but following the omitted card-free-trial route exposed another P1 boundary defect:

**[P1] A card-free calendar-yearly trial ending on the local first can charge the annual price twice.** `TryResolveTrialConversion` correctly returns a whole 30/30 or 31/31 first period. Renewal then always replaces the annual `period` with that one-month `stub`, even when `stub.IsProrated` is false, and always builds `convertingAnnual`. Because the fraction is whole, `PricingBasis` does not switch to the monthly stub basis: the conversion charge is already the full annual amount. It then also creates a pending annual period for the same year. `AtCheckout` adds that pending annual amount to the conversion charge immediately; `AtBoundary` leaves it to be charged again at the following sweep. A trial ending exactly 1 September should have no stub and no pending year: it should open and charge one annual period from 1 September to 1 September next year.

Add end-to-end creation→conversion tests for a trial ending on the local first under both `AtCheckout` and `AtBoundary`, asserting one annual amount, `PendingAnnualPeriod == null`, the annual current-period dates, and the next renewal one year later. Build `convertingAnnual` only when the resolved opening period is actually partial; when it is whole, retain the schedule-derived annual period rather than replacing it with the one-month `CalendarFirstPeriod`.

**[P1] A one-period promotion is not consumed for a non-trial calendar-yearly signup on the local first.** There is no stub/pending record in that case, so checkout charges the discounted annual price, but activation's `StubConsumedDiscountPeriod` requires `InitialChargeProrated == true` and increments nothing. The next annual renewal therefore receives the same one-period promotion again. Preserve the legacy anniversary behavior if required, but a new calendar-yearly whole-period checkout must consume its annual promotion once. Add a signup-on-first activation assertion followed by the next annual renewal.

So: the reported double-consumption defect is closed, but these two local-first cases should be fixed before the billing-state PR merges. Invoice work remains correctly tracked by #315.

fahad0208:
Both fixed in `8264ae4`. They turned out to share a cause worth naming.

## 1 — Trial ending on the local first

Confirmed. `TryResolveTrialConversion` succeeds whenever a card-free calendar trial has ended, and the caller then overrode the period with the stub's dates **unconditionally**. For a trial ending on the first, `TryResolveFirstPeriod` correctly reports `IsProrated: false` — but nothing was reading that, so a whole year got truncated to the stub's month, charged a year's money for it, and left a second year pending behind it.

The override and the pending-year build are now both gated on `stub.IsProrated`. A trial ending on a boundary steps straight into the year the schedule already derived and holds nothing.

## 2 — Promotion duration on a whole calendar first period

Confirmed. `InitialChargeProrated && InitialChargeDiscountApplied` was a proxy that stopped being true: it was written when a stub was the *only* way a calendar price could open, so proration stood in for "calendar-aligned". A signup on the first broke the proxy and took the discount without spending a period.

The condition now asks what it means:

```csharp
subscription.InitialChargeDiscountApplied &&
CalendarBillingAlignment.IsCalendarAligned(subscription.Price)
```

**Anniversary keeps its exemption**, for the reason it always had — counting its first period would shorten every plan already sold on one. What changes is that a calendar-aligned whole first period now counts, monthly as well as yearly. That is new behaviour for calendar-monthly signups on the first, but calendar alignment ships in this PR and #307, so no subscription sold before it is affected.

## The common cause

Both were proxies standing in for a condition that used to be equivalent and no longer is — `converting` for "converting into a stub", `InitialChargeProrated` for "calendar-aligned". Adding the yearly whole-period case broke both equivalences at once. Each now tests the thing it actually depends on.

## One thing I changed beyond the findings

A whole-period conversion recorded no first-charge figures at all, because those fields were only written for a stub. A card-free trial leaves them unset at signup, so the conversion is the only place they can be filled in — leaving them null would have lost the tracing for that charge entirely. Conversions now record them whether the period is a stub or a whole year; a whole one simply records a fraction that is not partial.

## Tests

- **Trial ending on the first**, as a theory over both `AtBoundary` and `AtCheckout`: exactly one charge, the annual amount, a period running 1 October to 1 October, renewal a year out, and nothing pending.
- **The same conversion is not reported as prorated**, and does record its charge.
- **Activation**: a calendar whole first period spends one period; an anniversary one still spends none.

Two existing activation tests had fixtures that implied a calendar price without setting one — they passed under the old proxy and failed under the explicit condition, which is the check working. Fixtures corrected rather than the condition loosened.

## Verification

- Server: `0 Error(s)`. Full suite **2878 tests, 154 failures — unchanged from `dev`**: 150 Mongo-dependent `XUnitTest.Integration.*` plus the 4 pre-existing `SubscriptionSimulationGuard`/permission failures.
- Calendar suite: **113 passed, 0 failed**.
- Client: 253 passed; `type-check` and `build` clean.

Invoice completeness remains with **#315**.

fahad0208:
Verified `8264ae4` against the two latest findings. Both are closed.

- Whole trial conversion now distinguishes `convertingToStub = converting && stub.IsProrated`. A trial ending on the local first retains the schedule-derived annual period, charges one annual amount, creates no `PendingAnnualPeriod`, and schedules the next charge one year later. The theory covers both `AtBoundary` and `AtCheckout` using the subscription produced by the real creation service.
- Conversion now passes the whole 30/30 or 31/31 fraction into `ApplySuccessAsync`, so the first-charge amount and discount flag are recorded while proration fields remain absent/false.
- Activation now asks the actual condition: a promotion-reduced opening charge on a calendar-aligned price. This consumes one period for monthly or yearly calendar signup on the first while retaining the established anniversary exemption.
- The prior prepaid-boundary guard remains correct, so `AtCheckout` still consumes once at payment and zero when the paid year merely opens.

The judgement call to include calendar-monthly whole first periods is consistent with the rule and safely scoped to calendar alignment. It also avoids preserving the same proxy bug in a second cadence.

No new billing-state findings. PR #313 is ready from this review perspective; the complete feature remains blocked on invoice snapshots in #315.

PR #314, opened 2026-08-25
Application-owned subscription invoices and credit notes

fahad0208:
## Why

The payment provider's invoice was the only durable statement of what a subscriber had been charged. That put the record of our revenue inside somebody else's product — unreachable when they are down, gone the day we change processor, shaped by their template rather than ours. It also meant the **signup payment had no invoice at all**, because a hosted checkout composes none, so a customer's very first charge was the one they could not get a document for.

This makes the application the issuer. Stripe stays the processor.

## What is issued

`SubscriptionFinancialDocument` is append-only. Three types over two number series:

| Type | Issued for | Series |
| --- | --- | --- |
| `Invoice` | Every settled positive charge: initial checkout, card-free trial's conversion charge, renewal, paid plan change, paid quantity increase, metered overage | `INV-{year}-{000001}` |
| `TrialInvoice` | Every trial start, card or no card. Zero total — it states the terms of a period nobody was charged for | The same `INV-` series, so a subscriber can see they have every invoice |
| `CreditNote` | A confirmed refund, and a change that banked subscription credit | `CRN-{year}-{000001}` |

Nothing is issued for a failed, abandoned or pending attempt, an ordinary cancellation, or credit merely being *consumed* by a later invoice — that is a deduction on the invoice it paid for, and a second document would count the same value twice.

## The decisions worth reviewing

**Exactly-once is a unique index, not a lock.** Every document is keyed on the event that caused it — a payment id, a refund id, a subscription plus a trial instant, a change reference — derived from a durable identifier and never from a clock or a counter. A redelivered webhook, a retried work item, two workers racing and the recovery sweep all derive the same key, so the second attempt finds the first document instead of allocating another number and sending another email.

**Numbers have gaps, on purpose.** Allocation is an atomic `$inc` and happens *before* the insert, so a number taken by an attempt that then loses the duplicate race is abandoned rather than reused. A gap is a question an auditor can answer from the ledger; a reused number is two documents claiming to be the same one, which nothing can answer.

**The issuer derives rather than being told.** A settled charge already carries which subscription, which period, what came off before tax, what tax and what credit, and the order id says what kind of charge it was — so one method covers all six charge paths, and the recovery sweep can issue a document for a payment nobody scheduled because it reads the same payment and reaches the same answer. A design where each money path passed its own figures in would leave recovery with nothing to read.

The initial charge is the exception: hosted checkout records no breakdown, so it is **recomputed** from the terms it was sold on. Safe only because every input is snapshotted on the subscription — and reported as a single gross figure where the recomputation does not reconcile to the amount actually taken. A breakdown that does not add up to what the customer paid is worse than no breakdown, because it looks authoritative.

**Each discount source appears separately.** "CHF 100 came off" cannot be read back into "the yearly price gave 8% and the coupon gave nothing", and which of them it was is exactly what somebody reconciling a two-year-old invoice is asking. The split is *re-attributed* from the two rates and the price's authored combination rather than stored as two more numbers that could contradict the figure that was charged.

**A partial refund reverses proportionally, by largest remainder.** Three thirds of a hundred rounded down is ninety-nine, and that missing unit is the whole reason not to round each part on its own. The credit note's subtotal, tax and total reconcile against the invoice exactly. A full refund short-circuits, so returning everything reverses everything with no rounding at all.

**The PDF is the document.** Rendered from this module's own template, stored through the platform's storage driver, hashed with SHA-256. An issued PDF is never regenerated against a newer template, and the repository refuses the second write rather than relying on anybody remembering. Delivery is its own work type so a template that throws or a bucket that is down retries all day without re-entering the code that allocates numbers, and without the invoice looking unissued meanwhile.

**Money paths only announce, and announcing never throws.** By then the money has moved and the state transition is committed, so a scheduling write that fails costs a later document rather than a failed charge. Both document work types sit last in the queue: nothing about a document affects entitlement, and it must never delay a renewal that does.

**One document is not sweep-recoverable, and the README says so.** The credit note for a change that banks credit: the value is folded into `CreditBalanceMinor` the moment the change commits and the balance cannot say which change put it there. That one is issued inline, best effort, and logged loudly if it fails.

**Refunds reach this module only by polling.** A refund confirms inside the payment module, which must never depend on subscriptions, so nothing there can announce it and this side has to come and look. A deliberate cost of keeping the dependency one-directional.

## Billing profiles

New, and required before money moves. A paid subscription, a plan change or a quantity increase is refused with `subscription_billing_profile_incomplete` naming the missing fields; free plans, previews, quantity *decreases* and renewals never are — none of those is a moment a person can be asked to fill in a form, and a renewal that refused to charge over one would cost the subscriber their service. `RequireBillingProfile` turns it off for an installation mid-migration.

The address and tax id are deliberately optional: many subscribers are individuals with neither, and refusing them over a field their jurisdiction does not ask for would be a billing rule invented here.

Every value is copied onto each document at issue, so editing the profile changes what future documents say and nothing already sent. Contacts are recorded per user as people act, so a document can name who initiated a change *as they were then* — a worker renewal names `System renewal` and no user, because none acted.

## API

`GET /api/subscriptions/invoices` now answers from the ledger, filterable by subscription, type, status and issue-date range. Field names change: `documentId` replaces `paymentDetailId` as the identifier and `documentType` replaces `invoiceType`.

`GET /api/subscriptions/invoices/{documentId}/pdf` **also accepts a payment id**, so links handed out by the old history keep working; only a pre-migration payment falls through to the provider's stored copy, and that fallback is documented as deprecated. Bytes are served from the endpoint rather than as a storage or provider URL — either kind is a bearer token for the document that cannot be revoked by revoking the caller's access.

`GET`/`PUT /api/subscription-billing-profile` are new.

The old payment-derived history service is deleted. Its repository stays: the simulation harness reads it, and the recovery sweep now does too.

## UI

Two pages under Subscriptions: **Invoices** — filterable list, each row expandable into its lines, every discount source, the settlement's two sides where there is one, and the parties as the document states them; and **Billing profile** — the form, with the missing fields stated up front rather than discovered as a validation error at checkout. Downloads go through the authenticated client, because a plain link would arrive without authorization.

## Verification

- Server **2761 passing**. The 4 `SubscriptionSimulation*` permission-guard failures are pre-existing on `dev`.
- Client typecheck, lint and production build clean; **284 subscription tests** passing.
- The ledger's Mongo guarantees — both unique indexes, the atomic counter under 40 concurrent allocations, the write-once PDF — are covered by integration tests that need a `mongod`, so they run in `mongo-integration` rather than locally.
- Documentation updated: the integration guide gains billing-profile and financial-document sections, the new endpoints, and the new error codes; the module README gains the design rationale.

The partial-refund allocation had a real bug that its test caught: the gross was being allocated independently alongside the discounts, so gross-less-discounts did not equal net on the credit note. It now derives gross from the net plus the scaled discounts, which makes the invariant hold by construction.


fahad0208:
## Review follow-up — all eight findings addressed in `9cf5f50`

The review's framing was right and worth restating, because it drove the design: the ledger's unique indexes prove a document cannot be issued **twice**, and say nothing about whether it is issued **at all**. Five of the eight findings turned out to share one cause — the obligation to write a document lived only in the work queue, which is a separate database with no transaction shared with the money.

### The structural change

`SubscriptionDocumentSource`, appended to `SubscriptionDetail.PendingDocumentSources` **in the same write as the transition that caused it** — the pattern `SubscriptionOutboxEvent` already sets for an event and its publication. It solves two of the findings at once, and makes the other three tractable.

| # | Finding | Fix |
| --- | --- | --- |
| P1 | Financial events can permanently miss document creation | Four recovery passes, none time-bounded. Recorded obligations (unbounded by construction — a partial index on `PendingDocumentSources.0` holds only the subscriptions currently owing something), settled charges, confirmed refunds, and **trials**, which had no sweep at all. The last three walk forward from marks in `SubscriptionDocumentCursors`. |
| P1 | Downgrade credit notes explicitly not durable | Appended inside the very compare-and-set that banks the credit — `TryChangePlanAsync` and `TryApplyQuantityChangeAsync` both take it. The credit and the record of what it owes commit together or neither does. No longer the one document a sweep cannot recover. |
| P1 | Recovery can describe an old payment using the current plan | The obligation freezes plan, price, quantities and period at the transition. The issuer prefers them and falls back to the live subscription only for events predating this — logging when it does, because that is the one case where a document can name the wrong plan. Figures still come off the payment. |
| P1 | PDF hash can disagree with the stored PDF | Storage key is the document id **plus the hash of the bytes**, and the bytes are written **before** the key is recorded. Two renders write two objects; the recorded key always has its file. |
| P1 | Merchant identity installation-wide | `SubscriptionMerchantProfile`, one per tenant, console-writable only. Configuration becomes the fallback, flagged `isInheritedFromConfiguration`. A tenant naming no seller blocks a user-initiated charge alongside the subscriber's own missing fields. |
| P2 | Initiating-person identity not captured | Taken from the authenticated context (`BlocksContext` carries a display name and email) and recorded when they act. It was copying the *organization's* billing contact, so every document said the finance mailbox had changed the plan whichever employee actually did. A user-started trial names them; `System renewal` is reserved for the clock, and a refund credit note says `System refund`. |
| P2 | Downgrade credit-note accounting incomplete | Decomposed from the **outgoing** side of the settlement at the moment of the change, while that plan's rate and mode are still in hand, reusing the partial-refund allocator. Linked to the invoice that charged for the period being adjusted. |
| P2 | Email delivery at-least-once | The publish is claimed before it happens; a later attempt republishes under the same `MessageId`, derived from the document id and carried in the mail's data context, and logs that it is repeating. Consumer-side dedup on that id is now a documented requirement. |

### Two things the review did not name, found on the way

**The worker's due-check had the same hole as the sweep.** It decided whether to schedule a document sweep by asking "has anything settled in the last six hours" — so an obligation older than the window never got a sweep scheduled for it. The sweep having no window is worth nothing if the thing that wakes it does. It now asks the unbounded question against the partial index, falling back to each sweep's own stored mark.

**Frozen quantity items were shared, not copied.** `[.. subscription.QuantityItems]` copies the list; the items are reference types, so the "frozen" snapshot followed the live seat count. Caught by a test asserting exactly the property the snapshot exists for.

### Also worth flagging

- `DocumentIssueLookbackHours` is replaced by `DocumentFirstPassReachDays` (400). It is used once per tenant, before that tenant has any mark, and bounds nothing ongoing.
- The merchant profile needed a UI or it could not be set: **Merchant profile** page added beside **Billing profile**.
- `PaymentExecutionContext` gains `UserName` and `Email`, read from the context it already reads. The payment module already derived `ActorId` from the email, so this stays within its remit.

### Verification

- Server **2806 passing**. The 4 `SubscriptionSimulation*` permission-guard failures are pre-existing on `dev`.
- Client typecheck, lint and production build clean; **289 subscription tests** passing.
- New unit tests for each property above, including the ones that were previously only asserted in prose: a late invoice naming the charged plan and seat count, a full batch not advancing its mark, a trial found with no obligation record, two renders not overwriting each other, a republished mail under one id, and the initiator being the person rather than the mailbox.
- New Mongo integration tests, because every one of these is about what the database does when two writers arrive at once or a process dies between two writes: the obligation riding the compare-and-set, **a losing version race leaving no obligation behind**, the windowless sweep query finding a three-year-old obligation, marks staying monotonic under 40 concurrent writes, and one seller per tenant.

PR #316, opened 2026-08-26
Collect a card up front when nothing is due today

fahad0208:
## What

A zero opening amount and a subscriber with no card on file used to be the same thing, because the only way to hold a card was to charge it. `requirePaymentMethodUpfront` separates them.

Signup now forks three ways:

| Opening amount | Card required | What happens |
| --- | --- | --- |
| more than zero | — | the existing payment checkout |
| zero | no | activates immediately, no `checkoutUrl` |
| zero | yes | a **card-setup** checkout; `Incomplete` until the card is stored |

A card is required when the amount is zero and either the plan sets `requirePaymentMethodUpfront`, or the subscription starts on a trial whose `trialRequiresPaymentMethod` is set. Both together is the combination the setting was asked for: genuinely free until the trial ends, with a card on file so the charge that ends it has something to bill.

## How

Stripe Checkout in `setup` mode — not a one-cent charge, which lands on a statement and has to be refunded, and not a zero-value PaymentIntent, which Stripe rejects. The SetupIntent it produces carries the off-session mandate the first renewal relies on.

`IPaymentMethodSetupService` is a near-twin of `HostedCheckoutInitiationService` and deliberately not a branch inside it: everything that path does about money — the preflight that refuses a zero amount, the currency conversion, the capture mode, the line item — is exactly what must not happen here. What it does share is the provider machinery, so a setup session is leased, recovered, resumed and confirmed by the code that already does those things.

It is internal on purpose. Collecting a card costs nothing and so has none of the natural limits a charge has, so `CreatePaymentMethodSetupRequest` is never bound from an HTTP body and no endpoint exposes it.

## The record it leaves is not a payment

A `PaymentDetail` under `PAYMENT_METHOD_SETUP` with a zero amount, because everything that tracks a hosted session already hangs off one: the initiation lease, the redirect URL, the webhook route, the stored-card write.

The hazard throughout is that it looks enough like a payment to be mistaken for one — it settles at `Authorized`, it carries a provider reference, it lives in the payments collection. So every reader that counts money excludes the flow explicitly: payment listings, refunds, captures. Invoice history already required `SUBSCRIPTION_INVOICE`, so it was structurally excluded. Activation does not record it as the opening charge — `InitialPaymentDetailId` stays null, because there is no charge and no invoice behind one. It never captures, so nothing that sums captured money picks it up.

## Two things behave differently from a charge

- **Failure is not fatal.** A declined charge ends the subscription; nothing was refused here, so it stays `Incomplete` and another attempt is free to succeed. The staleness sweep still expires it if nobody comes back.
- **An expired session is replaced.** A hosted session cannot be reopened and the provider would replay it under the key that opened it, so a retry mints a new one. `PaymentMethodSetupKeyFor` carries an attempt number, bumped by a compare-and-set on the number itself so two tabs retrying at once produce one session and the loser is told so. An expired *charge* is still a conflict: raising a second one is how the same money gets taken twice.

Cancelling while a setup is outstanding settles its link, so finishing the card form afterwards cannot start a subscription somebody has cancelled.

## One change with wider reach than the feature

`checkout.session.expired` and `payment_intent.canceled` normalise to `WebhookIntent.Cancelled`, which the transition switch dropped through `default` and ignored. They now reach the setup handler, which returns without writing anything for any other flow — so an abandoned checkout behaves exactly as it always has. Making it do otherwise is a decision about abandoned baskets, not one for a card-setup handler. An unroutable expiry is logged and dropped rather than retried to the dead-letter queue, for the same reason. Both are covered by tests.

## Also

- `requirePaymentMethodUpfront` runs end to end: plan entity, request, response, Mongo update, `PlanSnapshot` (so a later catalogue edit cannot rewrite what a subscriber was asked for), and the builder UI beside the trial's own card question.
- `ProviderConformanceTests` lists `IPaymentMethodSetupRequestFactory` as optional by design — a provider without one can still take every payment, and callers are told it cannot store a card rather than being handed a broken session.
- Guide and both module READMEs updated.

## Not in scope

The plan notes a zero-total period may still produce a zero-value application invoice if the financial-document rules require it. That comes from the period, not from the card, and nothing here changes it.

`customerEmail` is not passed through to prefill the Stripe form — the paid checkout path does not either, and matching it seemed better than diverging in one branch.

## Tests

`ZeroAmountCardSetupTests` (14) covers the three-way fork, a fully discounted first period under both settings, both trial shapes, the link purpose and key, resume, retry-after-expiry, the concurrent-retry gate, and that an expired charge is still a conflict. `PaymentMethodSetupTests` (10) covers the Stripe form shape, metadata reaching the SetupIntent, normalisation of both setup events, and the transition service including a late expiry and a foreign flow. Plus activation, cancellation, refund/capture preflight, plan mapping, and client mapping and builder tests.

- Server: 2912 tests. **4 failures, all pre-existing** (`SubscriptionSimulationGuard` / permission), unchanged from `dev`. 150 `XUnitTest.Integration.*` failures are the usual Mongo-less environment.
- Client: 258 passed, typecheck and build clean.

### DIFF
--- a/client/app/cross-modules/utilities/subscription-simulation/models/subscription-simulation.model.ts
+++ b/client/app/cross-modules/utilities/subscription-simulation/models/subscription-simulation.model.ts
@@ -63,6 +63,13 @@ export interface SimulatedSubscription {
   recurringAmountMinor: number;
   /** Only present while payment is outstanding; null once activated. */
   checkoutUrl: string | null;
+  /** Card-only checkout state; setup never represents money moving. */
+  pendingCheckout?: {
+    purpose: "PaymentMethodSetup";
+    state: "Pending" | "Failed" | "Expired";
+    errorCode: string | null;
+    checkoutUrl: string | null;
+  } | null;
   version: number;
 }
 
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/step-trial.test.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/step-trial.test.tsx
@@ -1,0 +1,60 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, expect, it } from "vitest";
+import { FormProvider, useForm } from "react-hook-form";
+
+import {
+  defaultSubscriptionPlanFormValues,
+  type CreateSubscriptionPlanFormValues,
+} from "../../schemas/subscription-plan.schema";
+import { StepTrial } from "./step-trial";
+
+const Harness = () => {
+  const form = useForm<CreateSubscriptionPlanFormValues>({
+    defaultValues: defaultSubscriptionPlanFormValues,
+  });
+
+  return (
+    <FormProvider {...form}>
+      <StepTrial />
+    </FormProvider>
+  );
+};
+
+const cardRequirement = () =>
+  screen.getByLabelText(
+    /Require a payment method before activation, even when nothing is due today/i,
+  );
+
+describe("requiring a card before activation", () => {
+  /**
+   * The setting is not about trials, and a plan with no trial is exactly the case it was added
+   * for — a free tier that will one day be billed. So it must be reachable without one.
+   */
+  it("is offered on a plan with no trial at all", () => {
+    render(<Harness />);
+
+    expect(cardRequirement()).toBeInTheDocument();
+    expect(cardRequirement()).not.toBeChecked();
+  });
+
+  it("explains what changes when it is on", async () => {
+    const user = userEvent.setup();
+    render(<Harness />);
+
+    expect(screen.getByText(/starts straight away/i)).toBeInTheDocument();
+
+    await user.click(cardRequirement());
+
+    expect(cardRequirement()).toBeChecked();
+    expect(screen.getByText(/card form that charges nothing/i)).toBeInTheDocument();
+  });
+
+  it("leaves the trial's own card question alone", async () => {
+    const user = userEvent.setup();
+    render(<Harness />);
+    await user.click(cardRequirement());
+
+    expect(screen.getByLabelText(/Require a card to start the trial/i)).toBeChecked();
+  });
+});
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/step-trial.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/step-trial.tsx
@@ -26,6 +26,10 @@ export const StepTrial = () => {
     control,
     name: "trialRequiresPaymentMethod",
   });
+  const requirePaymentMethodUpfront = useWatch({
+    control,
+    name: "requirePaymentMethodUpfront",
+  });
   const meters = useWatch({ control, name: "meters" });
 
   return (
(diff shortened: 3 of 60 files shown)


fahad0208:
I reviewed the current head `735b213` against the **Zero-Amount Signup with Card Required Up Front** plan. The plan/configuration plumbing, setup-mode Checkout request, retry attempt keys, cancellation CAS protection, and payment/refund/history exclusions are present. Two findings remain:

### [P1] Do not expose setup as confirmed before the stored card is durable and adopted

`PaymentMethodSetupWebhookStateTransitionService.ApplyAsync` first calls `ApplyAuthorisationAsync`, which writes `Authorized`, `WebhookConfirmedAtUtc`, and the payment outbox event. Only afterward does it call `ApplyAuthorisationTokenAsync` to persist the card. The subscription activation worker treats the first write alone as confirmation (`IsConfirmed`) and transitions the subscription before `AdoptProviderCustomerAsync` checks for the stored method.

Those operations are observed by independent workers, so activation can race between them. If token persistence fails or is merely slower, the subscription becomes Active/Trialing, adoption finds no method and only logs a warning, and there is no retry that later attaches the successfully stored card to the billing account. The next renewal can therefore fail despite setup having been reported successful.

This violates the planned success ordering: persist the stored payment method, then activate. Make card persistence/adoption part of the durable readiness condition—e.g. do not publish the setup-success state/outbox until the token is durable, and have activation require/adopt the exact stored method successfully before the subscription CAS (with a retryable failure instead of activation). Add a concurrency/failure test where activation runs after payment confirmation but before token persistence, and a retry test proving the billing account is eventually wired before access is granted.

Relevant code:
- `server/Payment.DomainService/Services/PaymentMethodSetupWebhookStateTransitionService.cs` (authorization is applied before the token lifecycle call)
- `server/Subscription.DomainService/Outbox/SubscriptionActivationProcessor.cs` (`IsConfirmed`, transition, then best-effort adoption)

### [P2] Expose failed/expired setup state on the current-subscription API

The plan explicitly requires failure or expiration to be exposed. `SubscriptionResponse` only exposes `status` and `checkoutUrl`; neither the setup payment status nor the link state/error is mapped. After the sweep abandons a failed setup, `GET /api/subscriptions/current` returns an `Incomplete` subscription with a null URL, which is indistinguishable from an expired session or a transiently missing link. Before that sweep, it can still return the stale redirect URL.

Add an explicit pending-checkout/setup projection (at minimum state such as `Pending | Failed | Expired`, plus a safe error code and usable URL only for Pending), derive it from the linked setup/payment record, and test both `setup_intent.setup_failed` and `checkout.session.expired` through `GET /api/subscriptions/current`.

Because P1 can grant access without a renewal-ready payment method, I would not merge this head yet.

fahad0208:
Both review findings are fixed in `5af4802`.

- Setup success now persists the stored card before payment confirmation/outbox publication, closing the activation race.
- Setup activation additionally requires the saved method to be attached to the billing account before granting access; a missing method/account leaves the subscription Incomplete and retryable.
- `GET /api/subscriptions/current` now exposes `pendingCheckout` with `purpose`, `state` (`Pending`, `Failed`, or `Expired`), safe `errorCode`, and a URL only while it is usable. The top-level `checkoutUrl` remains for compatibility.
- Terminal setup sessions are no longer returned as usable checkout URLs, so retry can create a fresh attempt.
- The guide and client response model are updated.

Verification: 46 focused server tests pass; full server build succeeds; client production build succeeds.

PR #317, opened 2026-08-26
fix(subscription): keep the organization scope across billing screens

fahad0208:
## Summary

Picking an organization on the plan list and then clicking through the sidebar silently dropped it. Billing profile went on to edit whichever organization the server resolved for an unscoped request — usually the console's own — and nothing on the screen said so, which is what made it confusing rather than merely wrong.

The scope now survives navigation, every screen says which organization it is showing, and a refusal points at the page that fixes it. The duplicate billing-contact inputs on the subscribe dialog are gone, and the organization's saved profile supplies the billing account's contact instead.

No migration. Existing billing accounts are brought up to date the next time the organization subscribes, rather than by a backfill — see **Reconciling an existing account** below.

## Navigation

- A shared subscription-link helper in two parts: one for menu entries, one for links built inside a page. Only paths under `/app/subscription` are touched — the scope is this module's idea and means nothing to Payment or IAM, so putting it on their links would advertise a filter they do not apply.
- `ScopedDashboardRoute` carries it onto the shell's sidebar. The sidebar renders above every page, so no page can reach the links that navigate away from it; this is the one place the menu tree and the current URL are both in hand. It covers desktop and mobile at once, since both render from the same menu tree.
- **The entry the page is already inside is left clean.** The shell marks an entry active with `pathname.startsWith(menu.path)`, and a query string appended there matches nothing — the active item would quietly stop being highlighted. Nothing is lost by it: the link left clean is the one that navigates nowhere.
- Three back links pointed at `/dashboard/subscription/plans`, **which is not a route in this app** — the real one is `/app/:itemId/subscription/plans`. Fixed through the helper, which gets the project segment and the organization right in one place.

## Billing profile

- Names the organization being edited and always shows the resolved id, derived from the response rather than the URL.
- Warns, and **withholds "Ready to be invoiced"**, when the server answered about a different organization than the link asked for. Naming another organization is honoured for the platform console alone; for anybody else the request is *ignored* rather than refused, so the page has to state what it loaded. A green light under the wrong company's name is how somebody fills in a form for the wrong organization.
- Caching was already keyed per organization, so it is unchanged — the defect was the derivation, not the cache.

## Failure handling

- The API's error `code` and `fields` now survive to the UI instead of being flattened to a message. A 400 arrives as an `HttpError` whose `errors` is the parsed body, so the envelope sits a level down; both shapes are read in one place rather than at each call site.
- `subscription_billing_profile_incomplete` renders as the page that fixes it, for that organization, rather than as JSON.
- **The field list is split in two.** The server folds `merchantLegalName` into the same list, and that one is the tenant's own seller identity. Sending a subscriber to add a legal name, when the missing name is the platform's own, sends them to correct something that was never theirs.

## Duplicate contact inputs

- Removed "Billing email" and "Billing name" from the subscribe dialog. They looked like the details an invoice needs and satisfied none of them: the server reads the organization's billing profile, while these went to the billing account. Two forms for one answer, free to disagree, and the one on screen was not the one that counted.
- The request fields stay for integrations that keep their own record of a customer. The server fills each **per field** from the profile's billing contact — a caller that sends an address and no name keeps the profile's name, because it meant the address and blanking the name would lose the only one there is.
- This decides where renewal and usage-threshold mail is sent and nothing else. What a document states about its recipient is still snapshotted when the document is issued, never read from here.
- Read through `ISubscriptionBillingProfileGuard` rather than injecting the repository a second time, for the reason that collaborator exists: the profile is one organization's answer to "who do we bill", and two services reading it two ways is how they come to disagree. Deliberately **not** gated on `RequireBillingProfile` — a free metered plan is never asked for a profile and still sends threshold mail, so the address is worth having wherever there is one.

## Reconciling an existing account

Added in `b0020dd`, from review feedback on this PR. Resolving the contact in the creation service was only half the job: `GetOrCreateAsync` returned an existing account **untouched**, so an organization that subscribed before filling its billing profile in kept the blank contact for good. Correcting the profile and subscribing again changed nothing, and renewal and usage-threshold mail went on going nowhere — or to the address the profile used to name. A billing account is one per organization and provider and outlives every subscription on it, so "create it correctly" was never enough.

- The operation is now `GetOrCreateAndReconcileAsync`, and does that. The old name described the half that worked, which is how the missing half went unnoticed; leaving a pure get-or-create beside it would have left the same footgun for the next caller, so it is renamed rather than supplemented.
- **One atomic upsert** keyed on the unique index, replacing a read-then-insert with a window between the two. Creation-only fields go under `$setOnInsert`, so a second signup cannot rewrite the id a subscription already points at or move the creation date.
- A null contact field **leaves what is stored alone rather than blanking it**, so a caller naming only an address cannot erase a name. A supplied value overwrites — that is what lets a corrected profile take effect. Explicit request values still take precedence over profile defaults, per field, resolved before the repository sees them.
- With nothing to reconcile the update is insert-only and an existing account is left exactly as it stands, timestamp included. Recording that nothing changed would be a lie a support conversation later has to unpick.

**One consequence worth a reviewer's agreement:** a profile-derived value now overwrites a stored one, so an integration that sets a contact once and later subscribes *without* naming it will see the profile's value take over. That is the direction the finding asked for — a stale address is the reported failure — and distinguishing "stored by an integration" from "stored by an older profile" would need provenance state nobody asked for. Documented in the interface, the README and the API docs: send the value on every request if you keep your own record of the customer.

**Why the tests missed it.** The repository was mocked to return the account it was handed — which is what an insert does, and not what an existing account does. The mock reported success for the write that never happened. That assumption is now named in the fixture, and the behaviour is pinned against a real collection instead.

## Verification

- **Server:** 2920 passing on this branch. 4 failures, all `SubscriptionSimulation*` permission-guard tests — verified failing identically on untouched `origin/dev`, so pre-existing. Build clean.
- **The four new reconcile tests are integration tests** and need a real MongoDB, so they run in CI's `mongo-integration` job rather than locally. Verified here that they compile and are discovered by the runner.
- **Client:** typecheck clean, lint clean on all 17 touched files, 337 subscription and layout tests passing, full suite 2136 passing. Five client test *files* fail to load on an SVG asset import in the devops/create-project area; none of them imports anything this PR touches.
- Both were re-run against the commit in an isolated worktree, not just the working tree.

New tests worth naming: the scope reaches subscription links and no others; the entry the page is inside stays clean so the sidebar keeps highlighting it; an organization id is escaped so it cannot forge a second parameter; the profile page withholds "ready to be invoiced" when the server answered about another organization; the envelope is read out of both a failing status and a 200; a missing seller identity links to the tenant's page and lists none of the subscriber's fields; the billing account takes the profile's contact, an integration's own values win, and a request naming only an address keeps the profile's name.

From the review finding: an account created without a contact takes one up while keeping its id and creation date; a contact that changed is followed; a field the caller does not know is left alone rather than blanked, and a call knowing neither touches nothing; sixteen concurrent signups converge on one account and one contact; and a subscription links to the account the repository *returned* rather than the one proposed — linking to an id that was never inserted would leave a renewal with no card to present.

## Two decisions worth a reviewer's eye

- **After saving, the page offers "Continue to plans" rather than redirecting.** Auto-navigating would stop somebody reviewing what they just saved and is hostile to anyone merely editing the profile. Easy to change if you would rather it were automatic.
- **Every subscription menu entry carries the scope, including merchant profile,** which ignores it. A menu rule with named exceptions is how the next entry added comes to be forgotten, and the benefit is that a detour through merchant profile does not silently reset the organization.

### DIFF
--- a/client/app/cross-modules/utilities/subscription-simulation/components/billing-profile-incomplete-notice.tsx
+++ b/client/app/cross-modules/utilities/subscription-simulation/components/billing-profile-incomplete-notice.tsx
@@ -1,0 +1,79 @@
+import { AlertCircle, ArrowRight } from "lucide-react";
+import { Link } from "react-router";
+import { Button } from "@/components/ui-kits/button/button";
+import { Card } from "@/components/ui-kits/card/card";
+import { useSubscriptionLink } from "../../subscription/hooks/use-subscription-link";
+import type { BillingProfileGap } from "../../subscription/utilities/subscription-api-failure";
+import { describeMissingProfileFields } from "../../subscription/utilities/financial-document-format";
+
+/**
+ * A refusal the reader can act on: what is missing, and the screen that collects it.
+ *
+ * This replaces printing the server's error envelope, which is what the dialogs used to do — a
+ * subscriber saw a field list in JSON and no indication that the fix was one page away, or which
+ * organization it was one page away for.
+ *
+ * The two halves are separated because they are somebody else's job each. The subscriber's own
+ * details sit on their billing profile; a missing seller identity is the tenant's own configuration,
+ * and a subscriber sent to fix that would be looking for a form that is not theirs.
+ */
+export const BillingProfileIncompleteNotice = ({
+  gap,
+  organizationId,
+}: {
+  gap: BillingProfileGap;
+  /**
+   * The organization the refused operation was for, so the link lands on the profile that was
+   * actually refused rather than on whichever one the URL happens to be showing.
+   */
+  organizationId: string | undefined;
+}) => {
+  const subscriptionLink = useSubscriptionLink();
+
+  return (
+    <Card
+      className="flex flex-col gap-3 border-amber-300 bg-amber-50 p-4"
+      data-testid="billing-profile-incomplete"
+    >
+      <div className="flex items-start gap-3">
+        <AlertCircle className="mt-0.5 h-5 w-5 shrink-0 text-amber-600" />
+        <div className="space-y-1 text-sm">
+          <p className="font-medium">This organization cannot be invoiced yet.</p>
+          <p className="text-muted-foreground">
+            A paid subscription is refused before anything is charged, because the invoice it would
+            owe has to name who it is addressed to.
+          </p>
+        </div>
+      </div>
+
+      {gap.subscriberFields.length > 0 && (
+        <div className="space-y-2 pl-8 text-sm">
+          <p className="text-muted-foreground" data-testid="billing-profile-missing">
+            {describeMissingProfileFields(gap.subscriberFields)}
+          </p>
+          <Button asChild size="sm" variant="outline">
+            <Link to={subscriptionLink("billing-profile", organizationId)}>
+              Complete the billing profile
+              <ArrowRight className="ml-2 h-4 w-4" />
+            </Link>
+          </Button>
+        </div>
+      )}
+
+      {gap.merchantMissing && (
+        <div className="space-y-2 pl-8 text-sm">
+          <p className="text-muted-foreground" data-testid="merchant-profile-missing">
+            This tenant has not named itself as the seller, and an invoice has to state who issued
+            it. That is set once, for the whole tenant, and not per organization.
+          </p>
+          <Button asChild size="sm" variant="outline">
+            <Link to={subscriptionLink("merchant-profile", null)}>
+              Name the seller
+              <ArrowRight className="ml-2 h-4 w-4" />
+            </Link>
+          </Button>
+        </div>
+      )}
+    </Card>
+  );
+};
--- a/client/app/cross-modules/utilities/subscription-simulation/components/change-plan-dialog.tsx
+++ b/client/app/cross-modules/utilities/subscription-simulation/components/change-plan-dialog.tsx
@@ -22,6 +22,12 @@ import {
 import { toast } from "@/hooks/use-toast";
 import type { SubscriptionPlan } from "../../subscription/models/subscription-plan.model";
 import { formatPrice } from "../../subscription/utilities/subscription-format";
+import {
+  billingProfileGapOf,
+  subscriptionApiFailure,
+  type BillingProfileGap,
+} from "../../subscription/utilities/subscription-api-failure";
+import { BillingProfileIncompleteNotice } from "./billing-profile-incomplete-notice";
 import { useChangeSubscriptionPlan } from "../hooks/use-change-subscription-plan";
 import type {
   SimulatedSubscription,
@@ -50,6 +56,7 @@ export const ChangePlanDialog = ({
   const [priceId, setPriceId] = useState("");
   const [quantities, setQuantities] = useState<Record<string, string>>({});
   const [formError, setFormError] = useState<string | null>(null);
+  const [profileGap, setProfileGap] = useState<BillingProfileGap | null>(null);
 
   const targetPlan = useMemo(
     () => plans.find((plan) => plan.planId === targetPlanId),
@@ -80,6 +87,7 @@ export const ChangePlanDialog = ({
 
   const submit = async () => {
     setFormError(null);
+    setProfileGap(null);
 
     if (!targetPlan || !priceId) {
       setFormError("Choose a target plan and price.");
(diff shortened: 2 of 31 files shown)


fahad0208:
I found one remaining gap against the billing-contact fallback part of the plan.\n\n**[P1] The saved profile does not repair an existing billing account**\n\nSubscriptionCreationService now correctly resolves the request/profile contact and passes it to GetOrCreateAsync. However, BillingAccountRepository.GetOrCreateAsync immediately returns an existing account unchanged. Therefore, an organization that already has a Stripe billing account with blank or stale BillingEmail/BillingName still keeps those values after completing its billing profile and subscribing again. Renewal/usage-threshold delivery continues to use the old account fields, so the UI fields were removed without reliably establishing their replacement.\n\nThis is easy to hit in the reported flow: the failed/abandoned earlier subscription may already have created the billing account before the user fixes the profile. The current unit tests mock GetOrCreateAsync as returning the proposed object, so they cannot detect the repository discarding the new contact.\n\nPlease make account contact reconciliation explicit and concurrency-safe (for example, a repository GetOrCreateAndUpdateContactAsync operation with the resolved contact and a defined precedence rule), and add a repository/integration test where an existing account has null/stale contact values. Also test that explicit request values still win over profile defaults.\n\nEverything else I checked aligns with the plan: subscription-only sidebar scoping, scoped back/post-save links, resolved-scope identity and mismatch warning, organization-keyed profile queries, structured API failures with a direct scoped repair link, removal of duplicate subscribe inputs, and unchanged request compatibility.

fahad0208:
Confirmed and fixed in `b0020dd`. You were right, and right about why the tests couldn't see it.

`GetOrCreateAsync` returned an existing account untouched, so resolving the contact in the creation service accomplished nothing for any organization that already had an account — which is every organization that subscribed before filling its profile in. That is precisely the case the feature exists for.

**What changed**

The operation is now `GetOrCreateAndReconcileAsync`. I renamed it rather than adding a second method beside it: the old name described the half that worked, which is how the missing half went unnoticed, and leaving a pure get-or-create in place would have left the same footgun for the next caller. One production call site, three integration call sites, five mocks.

It is a **single atomic upsert** keyed on the unique index, which also closes the read-then-insert window the old implementation had. Creation-only fields go under `$setOnInsert`, so a second signup cannot rewrite the id a subscription already points at or move the creation date.

On precedence, as you asked: explicit request values beat profile defaults **per field**, resolved before the repository sees them. A null leaves what is stored alone rather than blanking it, so a caller naming only an address cannot erase a name. With nothing to reconcile the update is insert-only and an existing account is left exactly as it stands — I deliberately did not touch its timestamp, since recording that nothing changed is a lie a support conversation later has to unpick.

**One consequence I want you to agree with rather than discover**

A profile-derived value now overwrites a stored one. That is the direction your finding asks for — a stale address is the reported failure — but it means an integration that sets a contact once and later subscribes *without* naming it will see the profile's value take over. Telling "stored by an integration" from "stored by an older profile" would need provenance state, which felt like inventing a rule nobody asked for. If you'd rather a profile default only filled blanks, say so and I'll add the provenance field — but note that would leave the stale case unfixed, which I think is the worse of the two.

It's documented in the interface XML, the module README and the API docs: send the value on every request if you keep your own record of the customer.

**Tests**

Your diagnosis of the gap was exact — the mock returned the account it was handed, which is what an insert does and not what an existing account does, so it reported success for the write that never happened. That assumption is now written down in the fixture, and the behaviour is pinned against a real collection instead:

- an account created without a contact takes one up, keeping its id and creation date;
- a contact that changed is followed;
- a field the caller does not know is left alone rather than blanked, and a call knowing neither field touches nothing;
- sixteen concurrent signups converge on one account and one contact.

Plus a unit test that a subscription links to the account the repository **returned** rather than the one proposed — linking to an id that was never inserted would leave a renewal with no card to present.

Those four are integration tests and need a real MongoDB, so they run in `mongo-integration` rather than locally; I verified here that they compile and are discovered by the runner. Server unit tests: 2920 passing, with the same 4 pre-existing `SubscriptionSimulation*` permission-guard failures I confirmed fail identically on untouched `origin/dev`.

I also corrected the PR description, which still claimed existing accounts keep their stored values — that was true when I wrote it and is exactly what this changed.

fahad0208:
Re-reviewed through 9f14ea.\n\nThe implementation now directly addresses my finding: creation resolves request/profile precedence per field, and the repository performs one atomic identity-keyed upsert that preserves the existing account id/provider state while reconciling non-null contact values. The added real-Mongo cases cover blank→filled, changed contact, partial contact, insert preservation, and concurrent creation. I found no additional code-level defect in this correction.\n\nIt is **not ready to merge yet**, for two objective reasons:\n\n1. GitHub currently reports **CONFLICTING** with dev. The conflicts include the subscribe dialog/service and SubscriptionCreationService, so they overlap the behavior under review rather than being harmless documentation conflicts.\n2. The latest subscription integration run is still the failed run on 1e89637 (17 passed, 2 failed). No integration workflow has run on current head 9f14ea, so the corrected serialized upsert and its two regression tests remain unverified against Mongo.\n\nI would **not close/reopen the PR first**. Rebase/merge current dev, resolve the five conflicting files carefully, and push. That push should both make the PR mergeable and emit the synchronize event needed for a fresh integration run. Only if the workflow still does not appear after that should it be manually re-run or the PR close/reopen workaround be used.\n\nVerdict: code-level finding resolved; approval remains blocked on conflict resolution and a green Mongo integration run at the final rebased SHA.

PR #318, opened 2026-08-26
feat(subscription): add configurable trial period rules

fahad0208:
## Summary

Replaces the trial's fixed "number of days" assumption with three
authoring choices:

- **Fixed number of days** (unchanged legacy behavior).
- **Until the end of the signup calendar month.**
- **Fixed number of anniversary months.**

Trial boundaries are resolved in the subscription's own configured time
zone and frozen onto `TrialTerms` at creation — a later catalogue edit to
the plan's trial rule can never move an existing subscriber's boundary.
Existing plans using the legacy `trialDays` are unaffected.

## API and data model

- New `TrialDurationKind` enum: `Days` (= 0, the default — so every
  document written before this field existed deserializes to exactly
  what it already meant) / `EndOfCalendarMonth` / `AnniversaryMonths`.
- `Plan` gains `TrialDurationKind`/`TrialDurationCount` alongside the
  legacy `TrialDays`. `TrialDurationNormalizer` is the single place that
  reconciles the two, used by both response mapping and subscription
  creation.
- `PlanDefinitionRequestValidator`: rejects a request naming both the
  legacy and current fields; enforces each kind's count rule (`Days`
  1-365, `AnniversaryMonths` 1-12, `EndOfCalendarMonth` none).
- `PlanResponse` always returns the normalized `trialDurationKind`/
  `trialDurationCount`, and keeps returning the legacy `trialDays` only
  for a `Days`-mode plan.

## Boundary and billing behavior

- `Days`: unchanged — ends exactly `count × 24 hours` after signup, no
  time zone involved.
- `AnniversaryMonths`: same local wall-clock time, N months later,
  clamped to the target month's last day (31 Jan + 1 month → 28/29 Feb)
  — the same clamp-without-writeback pattern `BillingPeriodCalculator`
  already uses for billing anchors.
- `EndOfCalendarMonth`: local midnight on the first of the next month.
  `EndsAtUtc` is an **exclusive** boundary — an August 31 signup can
  have a trial lasting only until September 1, by design.
- Both new modes go through the existing `BillingLocalTime` DST-gap and
  ambiguous-time handling, matching every other billing boundary in this
  module.
- A card-free trial's first paid period anchors on `Trial.EndsAtUtc`
  regardless of which kind produced it — verified for both anniversary
  and end-of-month modes (previously only proven for `Days`).

## Client

- `SubscriptionPlan`/`Create`/`UpdateSubscriptionPlanRequest` gain
  `trialDurationKind`/`trialDurationCount`. The console always authors
  through these now, never the legacy `trialDays`.
- The plan builder's "Trial length (days)" field is replaced with a
  duration-kind selector (including "No trial") and a conditional count
  input labeled "Days" or "Months", with a zod `superRefine` mirroring
  the server's validation rules exactly.
- Plan cards/summaries describe the selected rule instead of always
  saying "N-day trial".
- Reopening or duplicating a plan reads the server's already-normalized
  fields, so a legacy day-based plan reopens identically regardless of
  which format it was originally saved in.

## Docs

Added a "Trial duration" section to `Subscription.DomainService/README.md`
covering the three kinds, the exclusive-boundary semantics, time
zone/DST handling, the deliberate short-trial-on-a-late-signup behavior,
and worked examples for how each mode anchors the first post-trial
charge.

## Test plan

- [x] `dotnet build Api/Api.csproj --no-incremental`: 0 errors
- [x] `dotnet test XUnitTest` (non-integration): passes, no regressions
      (confirmed against a clean checkout: the few failures present are
      a pre-existing, unrelated `SubscriptionSimulationGuard`
      test-harness issue)
- [x] New `TrialDurationResolverTests`: all three kinds, a late-signup
      short trial, month-end clamping in a leap and non-leap year, a
      year-boundary crossing, and both a spring-forward-gap and an
      ambiguous-autumn-hour anniversary boundary (Europe/Zurich)
- [x] New `SubscriptionCreationServiceTests`,
      `PlanDefinitionRequestValidatorTests`, `PlanResponseMapperTests`
      cases for every duration kind and the legacy/current field
      conflict
- [x] `npx tsc -b --noEmit` / `npx vitest run` (client): 0 type errors,
      2118 tests passed (6 unrelated pre-existing suite failures: a
      canvas dependency, an unrelated parse error, a local tooling
      glitch — none touch subscription code)
- [x] `npx eslint` on every changed/new client file: 0 errors, 0
      warnings

## Follow-ups

If there is a separately-published integration guide beyond this repo's
own `Subscription.DomainService/README.md`, it still needs the same
exclusive-boundary/DST/post-trial-billing language added.

### DIFF
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx
@@ -134,7 +134,8 @@ const PlanBuilderWizard = ({
     displayName: draft.displayName ?? "",
     code: draft.code ?? "",
     organizationLabel,
-    trialDays: draft.trialDays ?? null,
+    trialDurationKind: draft.trialDurationKind ?? null,
+    trialDurationCount: draft.trialDurationCount ?? null,
     trialRequiresPaymentMethod: draft.trialRequiresPaymentMethod ?? true,
     quantityItems: (draft.quantityItems ?? []).map((item) => ({
       itemKey: item?.itemKey ?? "",
--- a/client/app/cross-modules/utilities/subscription/components/plan-builder/step-trial.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/plan-builder/step-trial.tsx
@@ -18,10 +18,36 @@ import {
 import type { CreateSubscriptionPlanFormValues } from "../../schemas/subscription-plan.schema";
 import { CardListItem, CardListShell } from "./card-list-shell";
 
+/** Not a server value — selecting this clears the trial fields entirely. */
+const NO_TRIAL = "None" as const;
+
+const TRIAL_DURATION_KIND_OPTIONS: {
+  value: "Days" | "EndOfCalendarMonth" | "AnniversaryMonths";
+  label: string;
+  description: string;
+}[] = [
+  {
+    value: "Days",
+    label: "Fixed number of days",
+    description: "14 days from signup.",
+  },
+  {
+    value: "EndOfCalendarMonth",
+    label: "Until the end of the calendar month",
+    description:
+      "Until the end of the calendar month; late signups may receive a short trial.",
+  },
+  {
+    value: "AnniversaryMonths",
+    label: "Fixed number of months",
+    description: "Same local date and time after N months.",
+  },
+];
+
 export const StepTrial = () => {
-  const { control } = useFormContext<CreateSubscriptionPlanFormValues>();
+  const { control, setValue } = useFormContext<CreateSubscriptionPlanFormValues>();
   const trialGrants = useFieldArray({ control, name: "trialGrants" });
-  const trialDays = useWatch({ control, name: "trialDays" });
+  const trialDurationKind = useWatch({ control, name: "trialDurationKind" });
   const trialRequiresPaymentMethod = useWatch({
     control,
     name: "trialRequiresPaymentMethod",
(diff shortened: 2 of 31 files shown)


fahad0208:
Addressed all three findings in 8fbdf4a:

**1. [P1] Edit persistence** — confirmed: `TryUpdatePlanAsync` builds its Mongo update from an explicit `.Set(...)` field list, and `TrialDurationKind`/`TrialDurationCount` were missing from it. Added both. New `SubscriptionCatalogueRepositoryIntegrationTests` asserts the persisted document after an edit, including the specific case you called out (a legacy day-based plan edited onto `EndOfCalendarMonth` — confirms the old count doesn't survive and the new kind does).

**2. [P1] Anniversary schedule anchor** — confirmed, and worth noting: this bug predates this PR (the anniversary branch already used `now` instead of `scheduleAnchorUtc` for the legacy day-based trial too — it just had no test asserting `FeeSchedule.AnchorInstantUtc` to catch it). Fixed by passing `scheduleAnchorUtc` into both branches. New tests assert the schedule anchor directly plus first *and* second paid-period boundaries, for an anniversary price under both new trial modes, and a calendar-aligned-price case confirming its own correct (calendar-snapped, not trial-end-snapped) anchor.

**3. [P2] Integration guide** — updated `server/Api/Documentation/subscription/v1/index.html` (the published client-facing guide) with the new fields, all three modes, exclusive-boundary phrasing, time-zone/DST handling, and post-trial billing anchoring for both price types. The README update from the first commit stays as internal documentation.

Full test plan and details in the commit message. One honest caveat: the new integration test couldn't be run locally (no mongod/docker available in this sandbox) — it's written against the same fixture pattern the existing `SubscriptionRepositoryIntegrationTests` already uses successfully, but hasn't been executed end-to-end by me.

PR #325, opened 2026-08-26
[#278] [FE] Sticky stepper and preview panel on the subscription plan builder

### LINKED ISSUE #278: [FE] Sticky stepper and preview panel on the subscription plan builder
# SPEC — Sticky stepper and preview panel on the subscription plan builder

**Target repo:** blocks-utilities

## 1. GOAL
On the subscription module's plan builder screen (used by both "Create subscription plan" and "Edit subscription plan"), the step progress bar at the top and the plan-preview panel on the right currently scroll away as soon as the user scrolls down through a step's form fields — losing the user's sense of progress and the live summary of what they're building. This change makes both persistently visible while scrolling: the stepper pins to the top of the viewport, and the preview panel keeps following in its column, positioned just below the stuck stepper. Priority: keeping both elements reliably visible while scrolling through long steps (e.g. "What the plan grants") outweighs preserving the exact current decorative visual treatment of the page background, which must be reworked (not removed) to stop blocking sticky positioning.

## 2. CONTEXT
- **Integrates with:**
  - `client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder.tsx` — the shared wizard shell; owns the `<main>` wrapper (currently `overflow-hidden`), the `PlanBuilderProgress` placement, and the `<aside>` preview column (`sticky top-5`, already present but non-functional).
  - `client/app/cross-modules/utilities/subscription/components/plan-builder/plan-builder-progress.tsx` — the stepper component itself; currently has no sticky positioning.
  - `client/app/cross-modules/utilities/subscription/pages/create-subscription-plan-page.tsx` and `edit-subscription-plan-page.tsx` — both render `PlanBuilder`, so both inherit this fix automatically since neither page defines its own layout.
- **Must NOT break:**
  - The decorative blurred background circles (two `absolute` divs in `plan-builder.tsx`) must remain visually clipped to the page — they must not bleed past the page edges once `overflow-hidden` is removed from `<main>`.
  - The existing responsive behavior: the preview `<aside>` stays `xl:` and up only; below `xl:` the collapsible "Preview this plan so far" accordion (unaffected by this change) remains the only preview affordance.
  - The stepper's existing mobile-collapsed rendering (numbered circles + `sm:hidden` current-step-title line) and its click-to-navigate behavior (`goToStep`, disabled for steps not yet reachable).
  - Any dropdown, select, popover, or combobox opened from a field inside the current step (e.g. organization picker, currency select in pricing/meter fields) must continue to render visually above the stepper and preview panel, not clipped or hidden beneath them.
- **Dependencies (other repos):** none. This is a self-contained frontend change to `blocks-utilities/client`; no other ecosystem repo is touched or relied upon.

## 3. DATA CONTRACTS
```
Input:  none (pure client-side layout/CSS change; no new props, no new form fields)
Output: none (no new data returned; existing PlanBuilder props/behavior unchanged)
Errors: none (no new failure modes introduced; this is a presentational change)
DB changes: none
```

## 4. EXAMPLES (at least 5)
- **1 (happy, stepper sticks):** User opens "Create subscription plan", lands on Step 3 "What the plan grants" (long list of meters via `CardListShell`), scrolls the page down 600px → the `PlanBuilderProgress` bar is now rendered pinned to the top of the viewport (`top: 0`), fully visible, still showing "Step 3 of 5" and the 5-node progress rail.
- **2 (happy, preview follows without overlap):** Same scroll as example 1, viewport width ≥ 1280px (`xl:` breakpoint) → the right-hand `PlanSummaryCard` preview is also stuck, positioned immediately below the stuck stepper's bottom edge with no visual overlap and no gap larger than the existing `space-y-5` rhythm (1.25rem).
- **3 (bad — dropdown must not be hidden):** User is on Step 1 "Identity", has scrolled so the stepper is stuck, and opens the organization `Select` dropdown → the dropdown's option list renders fully visible, layered above the stuck stepper and preview panel (no part of the open dropdown is clipped or covered).
- **4 (empty / first-run state):** User just opened the create-plan page with no scroll yet (`scrollY = 0`) → the stepper and preview render in their normal (non-stuck) document-flow position, pixel-identical to today's layout; no shadow/border treatment is shown since neither is stuck yet.
- **5 (boundary — narrow viewport, no preview column):** User resizes the browser to 1024px wide (below the `xl:` 1280px breakpoint) and scrolls down on Step 3 → the stepper still sticks to the top of the viewport (stepper stickiness is breakpoint-independent), but no preview `<aside>` is rendered at all (unchanged existing behavior — the collapsible accordion is the only preview affordance below `xl:`).

## 5. ACCEPTANCE CRITERIA — HAPPY PATH (at least 5, EARS, numbered, testable)
```
H1. WHEN the user scrolls the create- or edit-plan page down past the point where PlanBuilderProgress would leave the viewport, THE SYSTEM SHALL keep PlanBuilderProgress rendered fixed at the top of the viewport (top: 0) for the remainder of the scroll, at every viewport width.
H2. WHEN the viewport width is at or above the xl: breakpoint (1280px) AND the user scrolls past the point where the preview aside would leave the viewport, THE SYSTEM SHALL keep the PlanSummaryCard preview panel visible in its column, positioned so its top edge sits immediately below the stuck PlanBuilderProgress bar's bottom edge with no overlap.
H3. WHEN PlanBuilderProgress becomes stuck to the top of the viewport, THE SYSTEM SHALL apply a visible shadow/border treatment to it that is absent when it is in its normal (unstuck) document-flow position.
H4. WHEN the user opens a Select/dropdown/popover control from a field inside the current step while the stepper and/or preview panel are stuck, THE SYSTEM SHALL render that control's open content above both the stepper and the preview panel in stacking order.
H5. WHEN the page is scrolled back to the top (scrollY = 0), THE SYSTEM SHALL return PlanBuilderProgress and the preview panel to their normal in-flow position and remove the stuck-state shadow/border treatment.
H6. WHEN the create- or edit-plan page renders the two decorative background blur elements, THE SYSTEM SHALL keep them visually clipped within the page bounds (no bleed past the viewport edges) despite `<main>` no longer using `overflow-hidden`.
```

## 6. ACCEPTANCE CRITERIA — CRITICAL PATH (at least 5, EARS, numbered, testable)
```
C1. IF the viewport width is below the xl: breakpoint (below 1280px), THE SYSTEM SHALL still make PlanBuilderProgress sticky, while continuing to render no preview `<aside>` (unchanged from current below-xl: behavior).
C2. IF the user navigates between steps (Back/Next/click a reachable step node) while scrolled down and the stepper is stuck, THE SYSTEM SHALL keep the stepper stuck and update its "Step N of 5" label and active node in place, without any scroll-position jump caused by the sticky change itself.
C3. WHILE the mobile-collapsed "Preview this plan so far" accordion (below xl:) is expanded, THE SYSTEM SHALL leave that accordion's content in normal document flow (not sticky) — this change does not alter that affordance.
C4. IF a Select/dropdown/popover opened from a field inside the current step is left open while the user scrolls, THE SYSTEM SHALL keep that open control's content layered above the stepper and preview panel at all scroll positions, never clipped beneath either.
C5. IF the decorative background wrapper (introduced to replace `overflow-hidden` on `<main>`) is added, THE SYSTEM SHALL ensure it does not intercept pointer events on the form, stepper, or preview panel (the wrapper and its contents remain non-interactive).
C6. IF the browser window is resized across the xl: breakpoint while scrolled down with the stepper stuck, THE SYSTEM SHALL correctly show/hide the preview panel per the existing breakpoint rule without breaking the stepper's stuck state.
```

## 7. OUT OF SCOPE
- Any change to the collapsible "Preview this plan so far" accordion shown below the `xl:` breakpoint — it remains a non-sticky, in-flow accordion.
- Any change to the stepper's step-navigation logic, validation gating, or the set/order/titles of the 5 steps.
- Any change to `PlanSummaryCard`'s content or styling beyond what's needed for it to sit correctly below the stuck stepper.
- Any backend, API, or data-shape change — this is a frontend-only layout fix.
- End-to-end test instructions/coverage for this behavior (manual or automated) — excluded from this spec at the user's request.
- Sticky behavior for any other subscription-module screen (plan list, plan detail) — scoped strictly to the plan builder wizard (create + edit).

## ASSUMPTIONS / DECISIONS (confirm before building)
- A1. E2E test instructions are omitted from this spec, deviating from `SPEC_GUIDELINE.md`'s standard template (which normally requires them) — basis: explicit user instruction to exclude E2E from this spec.

## 8. VERIFICATION (definition of done)
- Automated tests written from every `H#` and `C#` above (component tests under `client/app/cross-modules/utilities/subscription/components/plan-builder/`, extending or adding alongside the existing `.test.tsx` files there), and all pass.
- Lint/build/test pass with the exact commands from the target repo's row in `ECOSYSTEM.md`:
  - `npm --prefix client run lint`
  - `npm --prefix client run type-check`
  - `npm --prefix client run build`
  - `npm --prefix client run test`
- No acceptance criterion unimplemented, nothing out-of-scope added, no unresolved questions remaining.
- Checklist: H1 done, H2 done, H3 done, H4 done, H5 done, H6 done, C1 done, C2 done, C3 done, C4 done, C5 done, C6 done.

rezwanx:
Closes #278.

Pins the plan builder's step progress bar to the top of the viewport while scrolling, and parks the plan-preview panel directly beneath it.

## The cause was one class, and it made the existing code inert

`overflow-hidden` on `<main>` made it the **scrolling box** for every sticky descendant — and `<main>` never scrolls, the page does. That is why the `sticky top-5` already present on the preview column did nothing: it wasn't mis-tuned, it was inert. Removing that clip is the whole fix; the decorative blurs are what it existed for, so they move into an absolutely-positioned clip layer that is **not** an ancestor of the sticky elements and therefore cannot re-break them.

**`overflow-clip` would have been the tidier one-word fix** and doesn't create a scroll container — I checked, and rejected it: it needs Safari 16, and this client declares no `browserslist` and no Vite build target, so Safari 14/15 are in the baseline. There it would silently stop clipping and the blurs would bleed past the page edge, violating H6 with nothing to show for it.

## Two things CSS can't express, in `use-sticky-stepper.ts`

**Is it stuck?** There is no `:stuck` selector, so the sticky element observes itself with the root inset 1px at the top. The obvious alternative — a sentinel above it — is worse in two independent ways, and both were found in review before implementation: a sentinel is *equally* non-intersecting when it sits **below** the viewport, so a short viewport or a restored scroll position reads as stuck; and a sentinel dropped into the `space-y-5` container becomes a **spacing participant**, shifting the very threshold it exists to detect. The ratio test is also paired with a top-edge check, because a ratio below 1 on its own likewise means "clipped by the bottom of a short viewport" or "wider/taller than the viewport under zoom".

**How tall is it?** The preview's offset must clear the stepper, and the stepper's height genuinely varies — its description line and step titles are `sm:`-only, its current-step title is `sm:hidden`. A hard-coded offset would overlap at some widths and gap at others, which is exactly what H2 rules out. So the border box is measured, and **ceiled** rather than rounded, because rounding a fractional height down leaves a sliver of overlap and H2 forbids overlap outright.

The stuck treatment is deliberately **dimension-neutral** — shadow and border colour, never width or padding. The measured height positions the preview, so changing the box on stick would move the preview every time the user crossed the threshold, and could provoke scroll anchoring.

## Verification, split honestly

**jsdom has no layout engine and implements neither observer**, so unit tests can only prove wiring. They do that thoroughly — 32 new tests covering the observer state machine, the offset arithmetic, the clip layer, the z-order, the `xl:` gating, and that the stepper's navigation and mobile rendering are untouched.

For the parts jsdom structurally cannot see, I ran the CSS mechanics in a real browser using a standalone harness mirroring this structure, and measured:

| criterion | measured |
|---|---|
| H1 stepper pins | at scrollY 600, `stepper.top = 0` ✓ |
| H2 preview below, no overlap | `preview.top − stepper.bottom = 20px` exactly; `overlap: no` ✓ |
| H2 measurement correct | measured height 81px = actual 81px ✓ |
| H6 blurs still clipped | blur extends to 1393px vs 1280px viewport, yet `document.scrollWidth = 1265 ≤ 1280` — no horizontal scroll ✓ |
| H4 portal layering | `elementFromPoint` over a `z-50` overlay covering the stuck `z-30` stepper returns the overlay ✓ |
| **H3/H5 stuck treatment** | **not verified** — see below |

**H3/H5 are unverified in a browser.** The harness tab reports `visibilityState=hidden`, and IntersectionObserver does not deliver entries to a non-rendered page, so the stuck-state detection never ran there. It is unit-tested, and the shadow is cosmetic rather than structural, but a reviewer scrolling the real page is the only thing that will actually confirm it. I'd rather say that than let a green suite imply otherwise.

I could not drive the real application: there is no `.env` (only `.env.example`) and reaching the plan builder needs an authenticated session, which I won't create.

**Recommended manual checks** a reviewer with a running app is well placed to make, from the code review: whether any ancestor between the wizard and the viewport sets a non-`visible` overflow in the real shell; whether a fixed global header or safe-area inset makes `top-0` tuck the stepper underneath it; behaviour at 200–400% zoom and in mobile landscape; and whether validation focus or `scrollIntoView` can now land a field beneath the stuck bar.

## Gates (§8, all four run individually)

| command | result |
|---|---|
| `npm --prefix client run type-check` | **clean** |
| `npm --prefix client run build` | **succeeds** |
| `npm --prefix client run test` | **2184 pass**; 5 file-level failures, all pre-existing |
| `npm --prefix client run lint` | **fails — pre-existing** |

Two of those need explaining rather than glossing:

- **`lint` already fails on `dev`**, independently of this change. I measured the baseline by stashing my work: **107 problems (33 errors, 74 warnings) both with and without it** — identical. My six files lint clean on their own (exit 0). Worth fixing separately; it is not fixable inside this ticket's scope.
- **5 test files fail to import**, also pre-existing (project-card, create-project ×2, repository-selection-modal, devops render-provider). Baseline is 5 failed files / 2152 passing; with this change it is 5 failed files / **2184** passing — exactly my 32 new tests and no change to the failures. One run also showed a flake in `environment-card` ("impersonates and navigates"); it passes in isolation and passed on re-run, and vitest isolates file environments so my observer stubs cannot reach it.

Also flagged: **§7 excludes E2E coverage from this spec** ("at the user's request", recorded as assumption A1) — so the browser evidence above is my own verification, not test coverage added to the repo.


fahad0208:
The failing gate came from an existing parse error in client/app/components/project-card/project-card.test.tsx. I updated the feature onto current dev, fixed that syntax error, removed the unrelated .gitignore delta, and verified 37 targeted tests plus type-check and production build. This PR's head is the protected inception branch, so GitHub rejected the update; replacement PR: #329.

PR #328, opened 2026-08-27
Expose and make subscription cancellation state idempotent

fahad0208:
## Summary

`GET /api/subscriptions/current` already returned `cancelAtPeriodEnd` and `canceledAtUtc`, but the contract was ambiguous about what was actually being told: whether cancellation was scheduled or already effective, and when access would actually stop. Repeating a cancellation request also had no defined behavior.

- Adds a nullable `cancellation` object to `SubscriptionResponse`: `state` (`Scheduled`/`Effective`), `requestedAtUtc`, `effectiveAtUtc` (period end while scheduled, actual end once effective), `canCancelImmediately`.
- Keeps `cancelAtPeriodEnd`/`canceledAtUtc` for backward compatibility.
- Persists whether a scheduled cancellation may be escalated immediately (`CanCancelImmediately`), computed once at scheduling time: `true` for an ordinary period-end cancellation, `false` when it's locked to an already-paid annual term. A record from before this field existed reads `false` — the conservative default, since escalating one without knowing whether it was annual-locked risks forfeiting paid-for access.

## Cancellation behavior

- Repeating `DELETE ...?immediately=false` after cancellation is already scheduled returns `200` with the existing state — no new transition, no version bump, original timestamps unchanged.
- `immediately=true` against an escalatable schedule ends it now (`state: "Effective"`).
- `immediately=true` against a schedule locked to a prepaid annual term stays idempotent and scheduled — the paid year isn't forfeited.
- A lost compare-and-set re-reads the subscription: if the winning write already produced the outcome this request wanted, it's reported as success (convergence), not a conflict. A genuinely different concurrent change still reports `subscription_transition_conflict`.
- `/current` continues returning a subscription with a scheduled cancellation — it's still live until `effectiveAtUtc` — and goes back to reporting no subscription once cancellation is effective (unchanged existing behavior, now covered by an explicit test).

## Design notes

- `EndAtPeriodEndAsync` now takes and persists `canCancelImmediately`; `Reflect(...)` mirrors it into the in-memory copy for the zero-extra-read response.
- New `SucceedAsync`/`ConvergeOrConflictAsync` helpers in `SubscriptionCancellationService` centralize the shared success bookkeeping and the CAS-loss convergence check respectively.
- No uncancel/resume endpoint added. Cancellation reasons remain internal. Existing response fields retained.

## Test plan

- [x] `dotnet build --no-incremental` — 0 errors
- [x] `dotnet test --filter "FullyQualifiedName!~Integration"` — only the 4 known pre-existing, unrelated `SubscriptionSimulation*` failures (verified independently on clean `dev`); all new/changed tests pass
- [x] New unit tests: escalation of an ordinary schedule, refusal to escalate a prepaid-annual-locked schedule, repeated period-end cancellation is write-free and version-stable, CAS-loss convergence on a matching concurrent write, CAS-loss conflict on a genuinely different state, `/current` mapping a persisted scheduled cancellation, mapper coverage for scheduled/effective/absent cancellation and legacy field back-compat
- [x] New integration tests (`SubscriptionRepositoryIntegrationTests`) for `CanCancelImmediately` persistence and concurrent duplicate-cancellation convergence against real Mongo compare-and-set — written but **not run locally**, no mongod available in this sandbox (consistent with every other integration test in the suite)
- [x] Published integration guide (`server/Api/Documentation/subscription/v1/index.html`) updated: idempotency rules, escalation rules, response examples, `/current` cross-reference


fahad0208:
Thanks for the thorough review — all three confirmed and fixed in 8576cbdd.

**1. Scheduled cancellation never enforced (blocking)** — Confirmed: nothing swept `CancelAtPeriodEnd` subscriptions past their `CurrentPeriodEndUtc`. Added `SubscriptionCancellationEffectiveProcessor` + `ISubscriptionRepository.ListDueForCancellationAsync` (mirrors `ListDueForRenewalAsync`, filtered on live status + `CancelAtPeriodEnd == true` + `CurrentPeriodEndUtc <= now`), wired into the existing work-scheduler machinery as `SubscriptionWorkType.CancellationEffective` (priority right after `Renewal`) with its own `CancellationEffectiveWorkHandler`, DI registration, and reconciliation due-check — same pattern as every other sweep in this module. On a hit it transitions to `Canceled`, sets `EndedAtUtc`, clears billing, invalidates the entitlement cache, and raises `SubscriptionCanceled` — the same transition shape as the interactive immediate-cancellation path.

**2. Concurrent cancellation requests can both write (blocking)** — Confirmed and root-caused exactly as you described: `EndAtPeriodEndAsync`'s transition doesn't move status, so two first-time requests both match the same `ExpectedStatus` filter. Added `SubscriptionTransition.RequireCancellationNotAlreadyScheduled`, enforced in `SubscriptionRepository.TryTransitionAsync` as an additional `CancelAtPeriodEnd == false` filter clause, set only on that one transition. Now exactly one write lands; the loser genuinely loses the CAS and goes through the existing re-read/converge path instead of also writing. Updated the Mongo integration test to actually exercise this (it was asserting the right behavior before but the transitions it built by hand didn't set the new flag, so it would have been testing nothing without this fix — good catch, thank you).

**3. Effective cancellation could still advertise `canCancelImmediately: true` (non-blocking, fixed)** — Confirmed: escalation left both `CancelAtPeriodEnd` and `CanCancelImmediately` from the prior schedule untouched. Both now clear to `false` on escalation, in `EndNowAsync`'s persisted transition and in `Reflect(...)`'s in-memory copy, so the immediate response and any subsequent read agree.

Also updated the integration guide with the sweep's existence/latency note, and added unit tests for all three (including a CAS-loss unit test for the new filter) plus two new Mongo integration tests (persisted-guard concurrency, sweep due-query). The integration tests are written but not run locally — no mongod in this sandbox, consistent with the rest of the suite; happy to have CI confirm.

Full build: 0 errors. Non-integration suite: only the 4 known pre-existing, unrelated `SubscriptionSimulation*` failures (verified independently on clean `dev`).

fahad0208:
Confirmed and fixed in 8217029c.

**Root cause**: both the scheduled-cancellation sweep and the interactive immediate/escalation path (`SubscriptionCancellationService.EndNowAsync`) moved status straight to `Canceled` without closing the still-open usage window first. The usage-rating sweep's due-query only ever matched live subscriptions, so once status flipped, nothing would ever look at that subscription's usage again — the final period's overage was silently forgone. (`EndNowAsync` actually carried an explicit "known, stated gap, not a built recovery path" comment for this — pre-existing, and now closed as part of this fix since the new escalation path shares the same code.)

**Fix — mirrors the existing plan-change pattern exactly**, rather than inventing a new mechanism:
- Plan changes already solve an identical problem: `TryChangePlanAsync` pushes a `PendingUsagePeriod` snapshot onto `PendingUsagePeriods` atomically with the schedule swap, and the usage-rating sweep prices whatever it finds there from the snapshot's own frozen plan/price terms.
- Added `SubscriptionTransition.OutgoingUsagePeriod`, set by both `EndNowAsync` (immediate cancellation and escalation) and `SubscriptionCancellationEffectiveProcessor` (the sweep), in the *same* atomic write that flips status to `Canceled` — so there's no window in which the status change can land without the usage window also being captured.
- Widened `ListDueForUsageRatingAsync`'s filter so a `Canceled` subscription still surfaces *only* while it holds a non-empty `PendingUsagePeriods` — otherwise it's excluded exactly as before.
- Added a guard in `SubscriptionUsageRatingProcessor.CloseSubscriptionAsync` so its live-clock loop (which advances `CurrentUsagePeriodStartUtc/EndUtc`) never runs for a subscription that isn't live — without this, a `Canceled` subscription found via its queued snapshot would also have its own current window double-processed by the second loop.

**Durability/crash-recovery**: comes from the same two guarantees the plan-change path already relies on — queuing is one atomic Mongo write with the status change (nothing to lose between "decide to cancel" and "queue the window"), and `EnsureInvoiceAsync`'s existing-invoice check (keyed by period key) makes rating idempotent no matter how many times the sweep revisits a queued period. Added a test simulating exactly that: an invoice already on file, confirming no duplicate is created but the pending-period pointer still clears.

**Tests added** (unit): immediate/escalated cancellation queues its final window; a `Canceled` subscription's queued window is rated with overage and rated as no-charge within allowance; the live-clock loop does not also advance a `Canceled` subscription's own window; a crash-retry (invoice already exists) doesn't duplicate the charge. **Tests added** (Mongo integration): a `Canceled` subscription with a queued window still surfaces from `ListDueForUsageRatingAsync`, one without does not — written but not run locally, no mongod in this sandbox.

Full build: 0 errors. Non-integration suite: only the 4 known pre-existing, unrelated `SubscriptionSimulation*` failures.

fahad0208:
Fixed in d529bf48.

`OutgoingUsagePeriodOf` now takes an explicit `effectiveAtUtc` and uses it as `PeriodEndUtc`, exactly as specified:

- **Immediate cancellation / escalation** (`SubscriptionCancellationService.EndNowAsync`): passes `now` — the instant the request itself stops entitlement.
- **Scheduled cancellation via the sweep** (`SubscriptionCancellationEffectiveProcessor`): passes `subscription.CurrentPeriodEndUtc` — the boundary that was *promised* while the cancellation was still "Scheduled" — not the worker's own execution time. `EndedAtUtc` is set to the same instant, for the same reason: a late pass must not silently stretch entitlement (or the invoice) to cover however long it was delayed.

`PeriodStartUtc`, the period key, and the frozen plan/price/currency snapshot are untouched, so a retry still resolves to the same invoice — the existing atomic-write + `EnsureInvoiceAsync` idempotency guarantees are unaffected by this change.

On "usage after the cancellation boundary excluded": no separate enforcement was needed. `UsageRecordingService` resolves the subscription through `GetLiveAsync`, which stops matching the instant status flips to `Canceled` — so no usage can be recorded against it at all once the transition lands, boundary-instant precision aside.

**Tests added**: immediate cancellation cutting a window midway through; escalation cutting a window midway through (distinct schedule-time vs. escalation-time instants); a late sweep pass retaining the promised boundary rather than its own run time; a usage window that naturally extends past its billing period being cut at the billing boundary instead of its own later end. Existing tests updated where they'd asserted the old (incorrect) `PeriodEndUtc == CurrentUsagePeriodEndUtc` behavior.

Full build: 0 errors. Non-integration suite: only the 4 known pre-existing, unrelated `SubscriptionSimulation*` failures.

fahad0208:
Implemented in c2dc59e5. This was the big one — summary below.

**Core fix**: liveness is now decided by `CancelAtPeriodEnd && CurrentPeriodEndUtc <= now`, not `Status` alone, everywhere it's checked:

- **Shared rule**: `Utilities/SubscriptionLiveness.IsEffectivelyLive(subscription, nowUtc)` for in-process checks, and the equivalent clause added to `SubscriptionRepository.BuildLiveFilter` for the Mongo query behind `GetLiveAsync`. `GetLiveAsync` now takes `nowUtc` as a **required** parameter (not defaulted) — a deliberately breaking signature change so no caller can silently keep the old, worker-dependent behavior. Updated all 4 production call sites (`UsageRecordingService` ×2, `SubscriptionCreationService`, `SubscriptionCheckoutService`, `EntitlementService`) and their test mocks.
- **Entitlement**: `EntitlementService` re-evaluates liveness against `nowUtc` on *every* read, including a cache hit — a snapshot cached a few seconds before the boundary does not go on granting for the rest of its TTL. `GrantingStatuses`/the old `Grants()` helper are gone, replaced by the shared predicate.
- **Usage recording**: `RecordAsync`/`GetCurrentUsageAsync` pass `nowUtc` into `GetLiveAsync`, and `ApplyAsync` revalidates immediately before the counter increment. I want to be upfront about scope here: this **narrows but does not fully close** the race, since nothing holds the subscription and the counter under one atomic write — documented explicitly in code and matching your spec's own allowed fallback ("document the remaining race"). What bounds the actual damage: the final window is frozen and rated exactly once regardless (from the earlier fix), so anything that slips through this narrowed window adds at most one increment to a period about to close, never an unbounded amount. A fully atomic close would need a reservation/claim or event-timestamp-filtered rating — flagging as follow-up scope if you want it closed further.

**Targeted scheduling**: `SubscriptionCancellationService` now schedules a per-subscription `CancellationEffective` work item (new `ISubscriptionWorkScheduler.ScheduleCancellationEffectiveAsync`) due at the same boundary just persisted, whenever a period-end cancellation is recorded — best effort and defensively try/caught, so a scheduling failure can't undo the cancellation itself. `CancellationEffectiveWorkHandler` gained the `AggregateId` branch (mirroring `RenewalWorkHandler`'s shape exactly), delegating to a new `ISubscriptionCancellationEffectiveProcessor.TryFinalizeAsync` extracted from the sweep's own per-subscription body — so the interactive-scheduled path and the tenant sweep can never drift onto different finalization rules. The tenant repair sweep (`ProcessDueAsync` / `ListDueForCancellationAsync`) is untouched and remains the durable recovery path regardless of whether the targeted item lands.

**Tests**: entitlement and `GetLiveAsync`-equivalent lookups stopping at the boundary while `Status` is still `Active`; a cached entitlement not outliving the boundary inside its own TTL; usage rejected after the boundary and accepted before it (with the counter increment verified as never called); targeted scheduling firing on a fresh schedule and correctly *not* firing on an immediate/escalated cancellation (nothing left to schedule); a scheduler failure not failing the cancellation; the work handler's `AggregateId` branch (found / not-found / already-escalated / not-yet-due). Two new Mongo integration tests for the boundary filter itself and for confirming an unscheduled subscription is unaffected by it.

Full build: 0 errors. Non-integration suite: only the 4 known pre-existing, unrelated `SubscriptionSimulation*` failures. `ServiceRegistrationTests` (DI graph resolution) passes, confirming the new constructor dependencies wire up cleanly.

fahad0208:
Implemented in efe1bf7c. This is the atomic mechanism the earlier liveness-recheck was explicitly acknowledged as only narrowing, not closing.

**Design**: two new Mongo collections, addressed by deterministic composite `ItemId` rather than a separate unique index (Mongo's own `_id` uniqueness is the enforcement):
- `UsagePeriodClosure` — one per `(tenant, subscription, periodKey)`: `Open` → `Closing` → `Closed`, `EffectiveEndUtc`, `ActiveWriterCount`.
- `UsagePeriodClaim` — one per `(tenant, subscription, periodKey, idempotencyKey)`, so a retried request reuses its original claim instead of double-counting.

**Wiring**:
- `UsageRecordingService.ApplyAsync` now acquires a claim before touching the ledger or counter. A rejected claim never reaches the ledger, the counter, or threshold evaluation. Released only by whichever call actually acquired the claim fresh — a concurrent duplicate that got `AlreadyClaimed` must not release on behalf of the still-in-flight original (I caught this exact bug while writing it: an early version let a fast duplicate release a slower original's claim mid-write). This replaces the earlier ad-hoc `SubscriptionLiveness` recheck rather than sitting alongside it.
- `SubscriptionCancellationService.EndNowAsync` and `SubscriptionCancellationEffectiveProcessor.TryFinalizeAsync` mark the period `Closing` — stopping new claims — before attempting the subscription transition. Best effort and defensively logged: entitlement stopping at `CurrentPeriodEndUtc` doesn't depend on this landing, only the tighter "no claim admitted even in the sub-second gap" guarantee does.
- `SubscriptionUsageRatingProcessor`'s `PendingUsagePeriods` loop checks `ActiveWriterCount` before rating; nonzero leaves the snapshot for the next sweep pass rather than a bespoke new retry-scheduling type, since the sweep already runs periodically.

**Tests**: unit tests for claim acquisition/rejection/release semantics (including the concurrent-duplicate-release bug above), the rating gate skipping a period with an active writer and proceeding once it's zero, and `StartClosingAsync` firing on immediate/escalation but not on scheduling. Mongo integration tests for the repository itself: concurrent claims all landing correctly, idempotent retry not double-counting, rejection once `Closing` or past the boundary, and `StartClosingAsync`/`TryMarkClosedAsync` idempotency.

**Scope, stated plainly, matching how the earlier narrowing was handled**: this does **not** implement a stale-claim reconciliation sweep (spec point 7) — a claim left `Active` by a crashed process isn't yet detected or force-released by anything. That's safe (it can never cause a period to rate too early) but would leave `ActiveWriterCount` stuck above zero for that one period until an operator or a follow-up sweep notices. I also didn't write all seven crash-recovery-per-transition-point scenarios as individual tests — the mechanism's crash-safety comes from every step being independently idempotent (claim insert, conditional increment/decrement, `Closing`/`Closed` transitions, `EnsureInvoiceAsync`'s existing-invoice check), which the tests above exercise but don't enumerate exhaustively. Happy to build the reconciliation sweep as a follow-up if you want it in this PR rather than after.

Full build: 0 errors. Non-integration suite: only the 4 known pre-existing, unrelated `SubscriptionSimulation*` failures. DI graph resolution (`ServiceRegistrationTests`) passes.

fahad0208:
Implemented the highest-value part of this in c2dc59e5..6c3885aa, plus resolved a real merge conflict against \`dev\`. I want to be upfront that I did **not** build the full spec as written — sections 4–7 (a four-stage leased claim state machine with a dedicated reconciliation worker and metrics) are a genuinely separate body of work, and I stopped short of building that blind. Here's exactly what's in and what isn't.

**What's fixed — the real gap in the previous round:**

The single-step `StartClosingAsync` had a bug your point 3 correctly identified: if the subscription transition that followed it lost its own compare-and-set to an *unrelated* write (a plan change, a quantity change — not the same cancellation), the closure record was left stuck `Closing` forever, and a subscription that was never actually cancelled would refuse ordinary usage indefinitely. Nothing released it.

Fixed with a proper reserve → commit/rollback protocol:
- `UsagePeriodClosureState` gains `CloseReserved`, between `Open` and `Closing`. Reserving blocks a *different* cancellation outcome from reserving the same period, but does **not** by itself stop ordinary usage claims — those still check `State in [Open, CloseReserved]` plus the boundary. Only committing to `Closing` (once the transition actually won) stops new claims; a reservation whose transition lost is released back to `Open`.
- `CloseOperationId` is now the deterministic `cancellation-close:{subscriptionId}:{effectiveAtUtcTicks}` your spec calls for, not the request's correlation id — so two writers racing to finalize the *same* intended cancellation (interactive escalation vs. the sweep) converge on one reservation instead of conflicting; only a genuinely different boundary is refused as `ConflictingOperation`.
- A reservation storage failure is no longer caught-and-logged — it propagates, so a cancellation can't proceed behind a reservation that was never actually taken (your point 2's "do not log and continue").
- The rating gate is tightened: a closure record that exists but hasn't reached `Closing` (still `CloseReserved`, or from a reservation that never committed) blocks rating exactly like an active writer would — scoped to periods that actually *have* a closure record, so a plan change's own outgoing window (which never creates one) is unaffected.
- Abandoned `Incomplete` checkouts skip closure/`PendingUsagePeriod` work entirely — they never activated, so there's no usage window to close.

**Also fixed: a real merge conflict.** `dev` moved with a large refactor (`SubscriptionReconciliationBackgroundService`'s dual queue/execute mode was removed entirely in favor of an announce-only `SubscriptionRepairAnnouncer`), which conflicted with this branch's earlier `CancellationEffective` due-check. Resolved by taking `dev`'s rewrite and re-adding the due-check in its new home; fixed the resulting test fallout (an unconfigured mock). Verified with `git merge-tree` against current `origin/dev` — no conflicts remain.

**What's explicitly not done, and why:** the 4-stage leased claim workflow (`Acquired → RecordWritten → CounterApplied → Completed`, lease ownership/expiry, a dedicated reconciliation background worker with metrics) is a real architectural addition on top of everything already in this PR, not a bug fix on top of it. The existing 2-state claim model (`Active`/`Released`) plus the reservation protocol above already prevents the specific "cancellation rates a period while usage is mid-write" race end-to-end for the cancellation side; what it doesn't yet give you is automatic recovery from a process that crashes mid-claim (a claim left `Active` forever by a dead process blocks that one period from ever reaching zero writers — safe, since it can't rate early, but stuck until someone notices). Building that properly deserves its own review rather than being bolted on at the end of an already-large diff. Happy to take it as a follow-up PR if you want it.

I also did not write the full crash-recovery-per-transition-point test matrix (retry after every one of the seven listed stages) for the same reason — the stages that exist today (claim insert, conditional increment/decrement, reservation commit/release, `Closing`/`Closed` transitions, `EnsureInvoiceAsync`'s existing-invoice check) are each independently idempotent and covered, but not enumerated one-crash-point-per-test.

Full build: 0 errors, including after the merge. Non-integration suite: only the 4 known pre-existing, unrelated `SubscriptionSimulation*` failures. DI graph resolution (`ServiceRegistrationTests`) passes post-merge.

fahad0208:
Pushed a follow-up round addressing the reviewer's two crash windows for cancellation idempotency (commit `fb5490ad`).

## What this closes

**1. Closure reservation stuck at `CloseReserved` (crash window 1).** If a cancellation's subscription transition committed but the follow-up `TryCommitClosingAsync` call itself never ran or failed, the closure was left `CloseReserved` forever, blocking final invoicing.

Fixed with a stale-reservation reconciliation: `IUsagePeriodClosureRepository.ListStaleReservationsAsync` finds `CloseReserved` closures older than a new `UsageClosureReservationTimeoutSeconds` (default 60s), and `SubscriptionCancellationService.ReconcileStaleClosuresAsync` resolves each one — validating the reservation's `CloseOperationId` against the deterministic `cancellation-close:{subscriptionId}:{effectiveEndUtcTicks}` shape before touching anything, then:
- committing to `Closing` if the subscription actually ended at the recorded boundary,
- releasing back to `Open` if the subscription is still live with no matching cancellation,
- leaving it alone and logging a warning if ambiguous.

Wired directly into `SubscriptionRepairAnnouncer.AnnounceAsync` (called inline, not via a new queued work type) — reconciling a stuck closure state moves no money and changes no subscriber's bill, so it doesn't need the sweep's own "never executes financial work directly" discipline, and a dedicated queue handler would only add a hop for something this cheap.

`TryCommitClosingAsync` / `TryReleaseReservationAsync` now return explicit `ClosureCommitOutcome` / `ClosureReleaseOutcome` enums (`Committed`/`AlreadyCommitted`/`OperationMismatch`/`StateConflict`/`NotFound`, and the release equivalents) instead of `Task`, and both `SubscriptionCancellationService.EndNowAsync` and `SubscriptionCancellationEffectiveProcessor.TryFinalizeAsync` now log a warning on a non-`Committed`/`AlreadyCommitted` outcome — the reconciliation above is what actually recovers it.

**2. Claim released without its counter decrement landing, or vice versa (crash window 2).** `UsagePeriodClaimState` gained a `ReleasePending` value (appended, not inserted, since the enum persists as an int). Release is now a resumable three-step protocol: CAS `Active → ReleasePending`, apply the `ActiveWriterCount` decrement idempotently, CAS `ReleasePending → Released`. The decrement uses a stable `claim-release:{claimId}` operation id tracked in a new `AppliedReleaseOperationIds` array on the closure document (a single filtered Mongo update: `NOT array-contains id AND ActiveWriterCount > 0`, combining `$inc` and `$addToSet`), so a retry after a crash resumes from the decrement rather than re-running it or skipping it.

The usage rating gate (`SubscriptionUsageRatingProcessor`) now also checks a new `HasOutstandingClaimsAsync` (any `Active` or `ReleasePending` claim for the period) alongside `ActiveWriterCount == 0`, so a claim stuck mid-release blocks rating even if the counter already reads zero. It logs a warning if the two signals disagree.

## Tests
Representative coverage added (Moq-style unit tests + Mongo integration tests, following the existing files' conventions):
- `SubscriptionCancellationServiceTests`: stale `CloseReserved` reconciliation committing for an actually-canceled subscription, releasing for a still-live one, leaving a mismatched-operation-id reservation untouched, and a no-op when no closure repository is configured.
- `SubscriptionUsageRatingProcessorTests`: rating refuses when a claim is `ReleasePending` even though `ActiveWriterCount` is already 0.
- `UsagePeriodClosureRepositoryIntegrationTests`: `TryCommitClosingAsync`/`TryReleaseReservationAsync` outcome enum correctness (`AlreadyCommitted`, `OperationMismatch`, `NotFound`, `AlreadyReleased`), the stale-reservation query's age filtering, a claim release resuming correctly from `ReleasePending`, and a retried release not double-decrementing.

Build is clean (0 errors) and `dotnet test --filter "FullyQualifiedName!~Integration"` passes except the same ~4 pre-existing `SubscriptionSimulation*` failures already verified unrelated to this branch. Integration tests require a live mongod not available in this sandbox, so they're written but unexecuted here.

## Deliberately deferred (as in prior rounds of this review)

Per the precedent set earlier in this cycle, I implemented the real, load-bearing fix for both named crash windows and am being explicit about what's left out of scope for this round:

- **The full stale-*claim* reconciliation sweep** (inspecting the usage ledger and counter operations per `Active` claim) — this needs a precise mapping against this codebase's actual usage-ledger entities that isn't specified closely enough to build safely in this pass, and is a genuinely separate feature. The claim-release resumability above already closes the specific crash window named in the ask (decrement/state desync); a claim that's merely abandoned mid-`Active` (never even started releasing) is not yet auto-recovered.
- **The full structured observability/metrics catalogue** (12+ named metrics) — only warning-level logs were added at the points that matter (commit/release outcome, signal disagreement, ambiguous reconciliation).
- **The exhaustive "pause after each of 10 named stages" crash-recovery test matrix** — representative tests cover the actual recovery paths instead.
- **Recoverable invoice-cleanup sequencing** — checked `SubscriptionUsageRatingProcessor`'s post-rating sequence (ensure invoice → schedule charge → remove `PendingUsagePeriod` → mark closure `Closed`) and it's already structured as independently-retryable steps (each guarded by its own idempotency: `TryCreateAsync`'s uniqueness index, `TryRemovePendingUsagePeriodAsync`, `TryMarkClosedAsync`), so no change was needed there.

Happy to open a follow-up PR for the stale-claim ledger reconciliation sweep and the full metrics catalogue if that's wanted — just say the word.

fahad0208:
Implemented and pushed as 63ab361b.

This closes the remaining recovery gaps instead of leaving them as deferred follow-up work:

- stale Active and ReleasePending usage claims are discovered and reconciled by the existing repair sweep, with conditional ownership before recovery;
- claim timestamps and a recovery index make that scan bounded and configurable;
- normal claim release removes its applied-operation marker, so AppliedReleaseOperationIds no longer grows once per usage request;
- an abandoned immediate-cancellation escalation is released even after the original scheduled cancellation boundary has arrived;
- final usage rating reads the append-first usage ledger, using counters only as a legacy fallback, so a crash after ledger append but before counter projection cannot lose billable overage.

Verification:
- focused cancellation/rating tests: 65 passed;
- full non-integration suite: 3030 passed, 1 skipped, only the same 4 pre-existing SubscriptionSimulation failures;
- build clean and git diff check clean.

CI/Mongo integration remains the final verification for the real-database paths.

PR #334, opened 2026-08-27
feat(subscription): brand invoice rendering with logo and colors

fahad0208:
Implements the **core** scope of the "Render Figma-Branded Invoices with Chromium" spec — the parts verifiable from this session (no Docker daemon, no browser-based visual diff available here). See **Explicitly out of scope** below for what's deliberately left for a follow-up.

## Design source

Figma node `1664:2` wasn't reachable from this session (no Figma tool exposed to Claude Code, and the browser pane here can't screenshot). Built from a screenshot/export the user provided directly.

## What changed

**Invoice template** (`FinancialDocumentHtmlTemplate.cs`) — rebuilt around the reference's sections: header (logo or text), seller/bill-to, a large colored headline, line items, totals, footer. Every existing data-driven section this app's document model actually carries is preserved (trial terms, settlement two-sides, prorated period, credit-note cross-reference) — just restyled into the new visual language. Two deliberate departures from the reference, both commented in the code:
- No due-date/hosted-payment-link concept exists in this system (everything is charged at issue) — the headline states what became of the charge (paid / credited / no payment due) instead of a due date.
- No per-line tax or bank-wire-detail field exists in the document model, so those sections aren't fabricated.

**Branding, tenant-configurable, snapshotted at issue**
- `SubscriptionMerchantProfile` gains `LogoFileId`, `PrimaryColor`, `AccentColor` (entity, request, response, validator — six-digit hex, normalized to `#` + uppercase).
- Copied onto `FinancialDocumentMerchant` at issue via the existing `ResolveAsync` path — no changes needed at the issuer call site.
- Caught and fixed the same class of bug flagged earlier this session: the repository's manual field-list `Update` builder needed the three new fields added explicitly, or they'd silently never persist.
- Unset or malformed color falls back to a shared default palette (`#17365D` / `#D9E7F5`) — both in the service (defence in depth) and again in the template itself.

**Safe logo embedding** (`IFinancialDocumentLogoResolver` / `FinancialDocumentLogoResolver`) — reads the snapshotted file once, up front (never during template rendering), validates PNG/JPEG/SVG by signature and a 512 KB limit, returns an already-embedded base64 data URI or nothing. The template keeps its "no network, ever" guarantee — a logo is either an embedded data URI or absent, never a URL. Missing/oversized/unrecognised bytes fall back to the merchant's name as text and log a structured, **non-blocking** warning (`document_logo_unavailable`) — branding never blocks a financial document.

**Delivery error codes split** — `document_pdf_unavailable` → `document_pdf_render_failed` / `document_pdf_storage_failed`, plus structured trace fields (tenant, document id/number, work-item id, attempt, stage, storage id) on every failure and success log line.

**Invoice list UI**
- States: `Preparing…` / `Download PDF` / `Generation failed`, driven by a new `IsAbandoned` field on the document response.
- Exposes the already-existing console-only resend operation as the recovery action for an abandoned document (this endpoint had no client surface at all before).
- Polls (`refetchInterval`) only while something on the visible page is genuinely still pending, stopping the moment it resolves either way.

**Merchant profile UI** — logo upload (reusing the app's existing pre-signed-upload pattern) with preview, and two color pickers (native `<input type="color">` + hex text field) wired to the new fields.

## Explicitly out of scope (flagged, not attempted)

- Chromium/Alpine installation in `Dockerfile.worker` and `PuppeteerSharp:ExecutablePath` configuration — confirmed via exploration that the worker image currently installs no Chromium and would fall back to `BrowserFetcher` downloading a glibc build that won't run on Alpine. This is a pre-existing gap, not introduced here, but blocks PDF rendering in production regardless of this PR.
- The startup/readiness probe that renders a minimal PDF.
- The container integration test.
- Visual regression against Figma reference images.

None of these are buildable or verifiable from this sandboxed session. Recommend as separate, focused follow-up work.

## Testing

- Server: `dotnet test --filter "FullyQualifiedName!~Integration"` — 2992 passed, only the known baseline failures (4 pre-existing `SubscriptionSimulation*`, 1 flaky unrelated `PeriodicPingBackgroundServiceTests`), 0 regressions. New: 7 template tests (logo embedding/fallback, color snapshot/fallback/malformed-defence, headline, footer), 3 merchant-profile round-trip tests (including the repository field-list catch), 2 logo-resolver tests.
- Client: `type-check`, `npm run build` (the exact command CI runs — this is what caught last time's interface-placement bug, so it's now part of my own verification loop), `eslint` on the whole subscription module — all clean. `vitest run` — 392/392 passed across the module, including new/updated tests on the invoices page (retry state, resend wiring) and merchant profile page (branding round-trip).

### DIFF
--- a/.github/workflows/worker-chromium-render.yml
+++ b/.github/workflows/worker-chromium-render.yml
@@ -1,0 +1,58 @@
+name: Worker Chromium render
+
+on:
+  pull_request:
+    paths:
+      - "Dockerfile.worker"
+      - "server/Utility.DomainService/PdfGenerator/**"
+      - "server/Worker/FinancialDocumentRendererReadinessCheck.cs"
+      - ".github/workflows/worker-chromium-render.yml"
+  push:
+    branches: [dev]
+    paths:
+      - "Dockerfile.worker"
+      - "server/Utility.DomainService/PdfGenerator/**"
+      - "server/Worker/FinancialDocumentRendererReadinessCheck.cs"
+
+jobs:
+  render:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Build the production worker image
+        run: docker build -f Dockerfile.worker -t blocks-worker-chromium-check .
+
+      # Not the C# render path itself -- that needs the full worker host (Mongo, secrets, a message
+      # bus) to start at all, which is a heavier fixture than this check is for. What this proves is
+      # the thing that was actually broken: whether the exact Chromium binary this image installs,
+      # under the exact launch flags PuppeteerSharpEngine passes, can render arbitrary HTML to a real
+      # PDF inside this container -- Alpine's musl libc, sandbox restrictions and all. Everything past
+      # that point is PuppeteerSharp driving this same binary over the DevTools protocol, which has no
+      # Alpine-specific failure mode once the binary itself launches. The command overrides the
+      # image's own ENTRYPOINT (dotnet Worker.dll) with a plain shell, since this is checking the
+      # container's installed Chromium, not starting the worker host.
+      - name: Render a real PDF with the image's installed Chromium
+        run: |
+          set -euo pipefail
+
+          docker run --rm --entrypoint sh blocks-worker-chromium-check -c '
+            set -e
+            chromium-browser \
+              --headless \
+              --no-sandbox \
+              --disable-setuid-sandbox \
+              --disable-dev-shm-usage \
+              --disable-gpu \
+              --disable-software-rasterizer \
+              --print-to-pdf=/tmp/render-check.pdf \
+              --print-to-pdf-no-header \
+              "data:text/html,<html><body>worker chromium render check</body></html>"
+            head -c 5 /tmp/render-check.pdf
+          ' > /tmp/pdf-head.txt
+
+          # %PDF- is the one thing that can only be true of a real, renderable document -- not of an
+          # empty file, an error page, or Chromium exiting non-zero with nothing written.
+          grep -q '%PDF-' /tmp/pdf-head.txt || { echo "No PDF was produced:"; cat /tmp/pdf-head.txt; exit 1; }
+          echo "Chromium rendered a real PDF inside the production worker image."
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -41,7 +41,31 @@ ENV DOTNET_ENVIRONMENT=Production \
 
 # icu-libs backs globalization; tzdata backs TimeZoneInfo. Alpine ships neither, and without
 # tzdata every FindSystemTimeZoneById throws here while passing on a developer machine.
-RUN apk add --no-cache icu-libs tzdata
+#
+# chromium and its runtime libraries back PuppeteerSharp's invoice PDF rendering. Installed here
+# rather than left to PuppeteerSharp's own BrowserFetcher, which downloads a glibc-linked Chromium
+# build at first use that does not run on this musl-libc base image at all — every render would
+# fail, silently, the first time this worker actually tried to issue a document. Alpine has shipped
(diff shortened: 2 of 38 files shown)


fahad0208:
Addressed all three findings, plus the Chromium deployment gap. New commit: `de18529c`.

## [P1] PDF hidden behind an unrelated mail failure — fixed

Reordered exactly as specified: `isPdfAvailable` always drives Download PDF, `isAbandoned` only matters when there's no PDF (a render/storage failure — safe to retry, nothing mailed). A PDF that exists but whose delivery abandoned on the mail side now shows a separate, **non-actionable** warning — no resend button from this page, worded differently for `document_no_recipient` (harmless) vs `document_mail_outcome_unknown` (needs a person). Added `lastErrorCode` to the response to drive that distinction. 4 new tests on `subscription-invoices-page.test.tsx` cover the exact scenario (PDF available + abandoned mail → Download PDF shown, no retry button, correct warning text per error code).

## [P1] mongo-integration — fixed and root-caused

Exactly as diagnosed: the integration fixture builds its own `ServiceCollection` by hand and never got the new registration. Added a deterministic `NoLogoResolver` fake (consistent with the fixture's existing `FixedMerchantProfile` naming no logo), and — since this exact class of bug already bit this PR once — added `IFinancialDocumentLogoResolver` to `SubscriptionDocumentRegistrationTests`'s descriptor list so a forgotten *production* registration would be caught there independent of any hand-built test fixture. Couldn't run `mongo-integration` myself (no local MongoDB), but the fix is a one-line registration matching an established pattern — CI will confirm.

## [P2] Logo validation — extracted and now actually tested

Pulled the byte-reading/signature-validation logic out of the resolver into `FinancialDocumentLogoBytesEmbedder` — pure, synchronous, no storage dependency, exactly as suggested. 11 new tests cover every path named in the review: valid PNG/JPEG/SVG, SVG with an XML prolog, unsupported signature, malformed SVG, empty bytes, exact-limit/one-over/multi-chunk streams, and cancellation mid-read. The resolver's own tests now cover only what only it can prove — storage returning nothing, storage throwing.

## Chromium in the worker image

This one I could verify for real, not just write and hope. Docker was actually available in this session (I'd wrongly assumed otherwise earlier) — I started the daemon, built `Dockerfile.worker`, and confirmed:
- The `chromium`/`chromium-browser` symlink resolves to a working binary (`Chromium 151.0.7922.173 Alpine Linux`).
- `PuppeteerSharp__ExecutablePath` is set correctly inside the image.
- Chromium actually renders a real PDF inside the built image, using the same launch flags `PuppeteerSharpEngine` passes (`%PDF-` magic bytes, non-empty, exit 0).

Added `.github/workflows/worker-chromium-render.yml`, which repeats that exact same check on every future change to the Dockerfile or the PDF engine — verified locally that the workflow's own shell command works before committing it, not just assumed. Also added `FinancialDocumentRendererReadinessCheck`, a hosted service registered first in the Worker's startup order that renders a probe PDF and throws if it can't — since this host has no HTTP surface for a `/health` endpoint (confirmed from `Program.cs` — plain `Host.CreateDefaultBuilder`, no Kestrel), a thrown exception from `StartAsync` is the only readiness signal available: it stops the Generic Host from starting at all, which is what turns a broken renderer into a container that never becomes ready instead of a worker that quietly claims and fails document-delivery work.

**What's still open:** visual regression against Figma reference images. That needs either Figma access (unavailable in this session) or the design team exporting and approving reference snapshots directly — a human dependency, not something I can close from here. The container check proves Chromium renders; it doesn't exercise the full C# worker host end-to-end (Mongo, secrets, message bus), which would need considerably more fixture than a smoke check justifies.

## Verified this round

- Server: `dotnet test --filter "FullyQualifiedName!~Integration"` — 3010 passed, only the known baseline (4 pre-existing `SubscriptionSimulation*`), 0 regressions.
- Client: `type-check`, `npm run build`, `eslint` — clean. `vitest run` — 395/395.
- Docker: real build, real render, verified by hand before committing the CI workflow that automates it.

PR #337, opened 2026-08-28
feat(subscription): author money in the currency's own units

fahad0208:
## Why

Two authoring fields asked for minor units while every other price on the same page asked for real money:

| Field | Was labelled | An author typing `5` meant | …and got |
|---|---|---|---|
| Meter overage tier | `Per unit (minor)` | 5 CHF per extra unit | **0.05 CHF** |
| Fixed discount | `Minor units off` | 5 CHF off | **0.05 CHF** |

Both are off by the currency's exponent, in the direction that quietly undercharges.

## What changed

Both fields now hold the currency's own units, and each converts at exactly one boundary — `plan-form-mapping` for tiers, the create call for discounts. `unitAmountMinor` and `amountMinor` are unchanged on the wire, and stored values are never rewritten.

**The exponent is the substance of this.** A minor unit is not always a hundredth, so the step, the placeholder and the permitted decimals are all read from the currency rather than assumed:

| | CHF / USD / EUR | JPY | BHD / KWD |
|---|---|---|---|
| Decimals | 2 | 0 | 3 |
| Input step | `0.01` | `1` | `0.001` |
| Placeholder | `0.05` | `5` | `0.005` |

A placeholder now never demonstrates an amount the same field would reject — there is no 0.05 in yen.

## Precision is rejected, not rounded

`toMinorUnits` rounds, which is right for arithmetic and wrong for authoring. New `isRepresentableInMinorUnits` is asked first, so `0.005 CHF` and `100.5 JPY` are named instead of becoming `0.01` and `101`.

Checked where both halves are in scope — on the rate table, not the amount field, since the allowance depends on the currency chosen beside it.

**This also closes the same hole on the plan price amount**, which was already in major units and already rounding silently: `89.999 CHF` was charged as `90.00` with nothing said. Outside the two fields named in the brief, but the identical defect one field over, so it seemed wrong to leave.

## Changing the currency

Nothing is converted — 0.05 francs is not 0.05 yen, and guessing which was meant would be inventing a price. The table is revalidated instead, so an amount the new currency cannot express is named.

The discount currency becomes a list rather than three free characters: an unrecognised code falls back to two decimals, which is wrong for yen and wrong by a factor of ten for dinars.

## Verification

- `tsc --noEmit` clean; `eslint` clean on all changed files.
- Subscription suite: **459 passed**. The only failures are 3 in `step-trial.test.tsx`, which I reproduced on clean `origin/dev` (`ab8b5adc`) before attributing them — its assertions look for copy the component no longer has.
- New `currency-precision.test.ts` and `meter-tier-units.test.ts` — **57 tests**, covering the brief's plan directly:

| Case | Asserted |
|---|---|
| `145.00 CHF` | → `14500` |
| `0.05 USD` | → `5` |
| `100 JPY` | → `100` |
| `1.250 KWD` | → `1250` |
| Stored values reopen | `14500 CHF` → `145`, `1250 KWD` → `1.25` |
| Round-trip | opening and saving unchanged does not move the stored figure |
| Invalid precision | rejected, with the currency's own limit named |
| Two tables, one meter | each priced by *its own* currency, not a shared exponent |
| `0.07 CHF` | accepted — held as `0.07000000000000001` after `Number()`, still seven centimes |

`describeDiscountAmountProblem` was extracted from the page so the discount rule is testable without rendering it.

No server change. No migration. No API property renamed.


fahad0208:
### Full client suite, and what the failures actually were

`npx vitest run` on `7592e559`: **2262 passed, 9 failed**, plus 4 files that fail to load.

Rather than reading "unrelated module" as "not my problem", I re-ran every failing file on its own:

| Files | Failures | Verdict |
|---|---|---|
| `payment-filters`, `create-payment-provider-page`, `payment-list-page`, `add-edit-provider-modal` ×2 | 6 | **Flaky under load** — all 60 tests pass in isolation. They were 5s timeouts recorded while a second vitest run was still finishing; the same files showed 20–25s durations. |
| `step-trial` | 3 | **Pre-existing on `dev`** — fails 3/3 on clean `ab8b5adc`. Its assertions look for copy the component no longer renders ("Require a payment method before activation…" vs. the current "Require a card to start the trial"). |
| `create-project`, `repository-selection-modal`, `create-project-resources-form`, `render-provider` | load error | **Pre-existing** — `file:///assets/github-icon.svg`, reproduced on clean `dev`. |

The subscription area on its own: **459 passed**, only the 3 `step-trial` failures above. `plan-price-alignment`, `plan-price-yearly-alignment`, `billing-alignment` and `subscription-price.schema` all pass (36 tests) — the files most exposed to the new price-amount precision rule.

Worth noting for anyone reviewing: `build-client` is skipped on this PR, as it is on every client-only change, so CI compiles and tests none of this. The evidence above is a local run.

fahad0208:
### [P1] fixed in `f2122d7e`

You were right, and the way it was wrong is worth naming: `1e-6` was dressed as a floating-point tolerance but is a business quantity — a hundredth of a centime. So the guard whose whole purpose was to stop silent rounding was licensing it, for every amount below that threshold. `0.000000001 CHF` validated, submitted as `0`, and gave the overage away; a fixed discount did the same and was refused by the server for a reason the author never saw.

```ts
const minor = majorAmount * 10 ** minorUnitExponent(currencyCode);
const rounded = Math.round(minor);
const tolerance = Number.EPSILON * Math.max(1, Math.abs(minor)) * 4;

return Number.isSafeInteger(rounded) && Math.abs(minor - rounded) <= tolerance;
```

Both parts of your correction, and the safe-integer bound earns its place independently: past 2^53 the gap between doubles exceeds one minor unit, so "is this a whole number of centimes" has no answer — and the scaled tolerance would then forgive several centimes at once. That is the failure mode a magnitude-scaled tolerance introduces, so it needed closing in the same change.

**The new cases were confirmed against the old implementation before the fix landed.** I reverted the tolerance, ran them, and got 5 failures pinpointing exactly this defect; restored, all 65 pass. Without that step they would have been tests of the implementation rather than of the bug.

Added:

| Case | Expected |
|---|---|
| `0.000000001 CHF`, `1e-9 JPY`, `0.0000004 KWD` | refused |
| `0.004 CHF` (just under half a centime) | refused |
| `MAX_SAFE_INTEGER`, `1e18` CHF, `1e300` JPY | refused — not a whole number of minor units |
| `1_000_000_000 CHF`, `1e15 JPY` | still accepted — absurd but expressible |
| Anything refused | asserted to be exactly what would have submitted as `0` |
| `describeDiscountAmountProblem("0.000000001", "CHF")` | names the 2-decimal limit |

Subscription suite: **467 passed**, 3 failures, all in `step-trial.test.tsx` and pre-existing on `dev`. Those are fixed in the CI PR below, where they had to be.

### CI

Your diagnosis is the correct one and mine was wrong — it is not a path filter. `build-client` in `ci-dev.yml` is gated on `github.event_name == 'push'`, so it skips every pull request by design; it publishes an image from `dev` rather than checking a change. The commented-out `pr-checks` job beside it was .NET-only, and `RUN_TESTS` is `"false"`. Raised separately as requested.

PR #338, opened 2026-08-28
Add server-owned metered overage preview endpoint

fahad0208:
## Summary

Adds `POST /api/subscription-usage/overage/preview` — a read-only endpoint that estimates the cost of additional metered usage using the active subscription's snapshotted terms and the *exact* rating, discount and tax logic used for the final usage invoice. The preview neither records usage nor charges the customer.

The core ask — a shared calculator so the preview and period-end invoicing can never drift apart — is implemented as two extractions:

- **`UsageChargeCalculator`** (`Subscription.DomainService/Services/UsageChargeCalculator.cs`): the discount+tax logic previously inlined in `SubscriptionUsageRatingProcessor.EnsureInvoiceAsync` is now `UsageChargeCalculator.Charge(grossAmountMinor, price)`, called by both the rating processor and the new preview service. Also provides `Difference(projected, baseline)` for the "additional = projected − current" rule below.
- **`SubscriptionUsageRater.OverageAllocations`**: the tier walker now reports per-band allocations over an arbitrary overage range (with an optional exclusive starting point, `fromOverageUnitsExclusive`), not just a total. `OverageAmountMinor` is kept as a compatibility wrapper so `SubscriptionUsageRatingProcessor` and any other existing caller are unaffected.

### New files (server)
- `Subscription.DomainService/Services/UsageChargeCalculator.cs`
- `Subscription.DomainService/Services/SubscriptionUsageOveragePreviewService.cs` (+ `ISubscriptionUsageOveragePreviewService.cs`)
- `Subscription.DomainService/Requests/PreviewUsageOverageRequest.cs` (+ `Validators/PreviewUsageOverageRequestValidator.cs`)
- `Subscription.DomainService/Responses/SubscriptionUsageOveragePreviewResponse.cs`
- `POST` action `PreviewOverage` on `Api/Controllers/SubscriptionUsageController.cs`

### New files (client)
- `use-preview-usage-overage.ts` hook (thin `useMutation` fetch wrapper — no client-side pricing)
- `PreviewUsageOverageRequest` / `UsageOveragePreviewResult` (+ nested) types and a `previewUsageOverage` service method in the existing `subscription-simulation` module

### Documentation
- New "Metered overage preview" section in `Subscription.DomainService/README.md`, matching the existing prose style (what it reuses, what it deliberately never does).

## Design notes

- **Ledger vs counter**: same precedence as period-end rating — the append-only ledger is authoritative whenever it has records; the counter is only a fallback for a legacy period with no ledger records.
- **Allowance**: resolved via `IMeterAllowanceResolver.EffectiveAsync`, so a trial grant or carried-forward allowance changes `includedQuantity` in the preview exactly as it changes enforcement.
- **`additionalCharge` is a difference, never rated standalone**: computed as `projectedPeriodCharge − currentCharge`, field by field, because a tier boundary or a rounding step at the discount/tax boundary can price the same units differently depending on what came before them.
- **Promotional discounts are explicitly excluded**, matching current behavior (only the price's own `AutomaticDiscountBasisPoints` reaches metered overage) — `discount.promotionalCodeApplied` is always `false` rather than silently omitted.
- **No writes possible by construction**: the preview service's constructor only takes read-capable dependencies (no invoice repository, no billing gateway, no outbox, no audit trail, no closure repository) — there is nothing here that could write even by mistake, not just nothing that does.

## Judgment calls / scope decisions

- **Route placement**: nested the action under the existing `SubscriptionUsageController` (`subscription-usage/overage/preview`) rather than a new controller, mirroring how `GetCurrent` already lives there.
- **Error code → HTTP mapping**: reused the existing `PaymentFailureKind` → status code switch (`SubscriptionApiResults.ToActionResult`) rather than inventing new plumbing — `Conflict` → 409 for both `subscription_meter_overage_not_allowed` and `subscription_meter_rate_unavailable`, `Unavailable` → 503 for `subscription_schedule_unavailable`.
- **Client module placement**: added the new hook/service method/types to the existing `client/app/cross-modules/utilities/subscription-simulation` module rather than a separate module, since that module already hosts the real (non-simulation) `/api/subscription-usage` client code (`recordUsage`, `RecordUsageRequest`) alongside the simulation harness — this endpoint follows that existing precedent rather than starting a new module for one endpoint.
- **`additionalTierBreakdown` is informational only**: it reports which tier bands the additional units crossed and what each contributed (for display), but is not itself re-summed into the charge — the authoritative `additionalCharge` is always the difference of two fully-rated totals, per the spec's explicit instruction not to independently tax/discount the new units alone.
- **No controller tests existed previously** for `SubscriptionUsageController` (only `GetCurrent`/`Record` were untested at the controller layer) — added a new `SubscriptionUsageControllerTests.cs` covering the new action only (auth attribute presence, envelope shape, validation-failure status/code, named-failure → status mapping, and organization-id forwarding), rather than backfilling coverage for the pre-existing actions, to stay scoped to this change.

## Test plan

- [x] Current usage 170, allowance 150, plus 100 → projected usage 270, projected overage 120
- [x] Additional usage crossing multiple tiers returns exact per-band units/amounts
- [x] Additional total == projected total − current total, including rounding at discount/tax boundaries
- [x] Automatic discounts work under both combination policies; promotional discount stays excluded
- [x] Inclusive and exclusive tax previews exactly match period-end invoice calculations
- [x] Trial grants and carry-forward allowance affect the previewed included quantity
- [x] Subscription snapshot terms are what's read (service never touches the plan catalogue — no catalogue dependency exists on the service at all)
- [x] Missing currency rate table and overage-not-allowed return named errors, not a misleading zero price
- [x] Ledger usage wins when a crash left the counter projection behind
- [x] Repositories verified to receive no mutation calls (`TryAppendRecordAsync`, `ApplyDeltaAsync`, `TryRepairCounterAsync`, `TryMarkThresholdNotifiedAsync`, `TryTransitionAsync` — all `Times.Never`)
- [x] Controller tests: authentication (`[Authorize]` presence), organization scope forwarding, response envelope, validation, correlation id
- [x] Parity test: same subscription + usage fed into preview and final rating (`SubscriptionUsageRatingProcessor`) asserts identical gross, discount, tax and total

## Build/test verification

- `dotnet build server/Api/Api.csproj` — 0 errors
- `dotnet test --filter "FullyQualifiedName!~Integration"` (from `server/`) — 3103 passed, 4 pre-existing failures in `SubscriptionSimulation*` tests confirmed unrelated (reproduced identically on unmodified `origin/dev` via `git stash`), 1 skipped
- `npm run type-check` (from `client/`) — 0 errors

### DIFF
--- a/client/app/cross-modules/utilities/subscription-simulation/constants/subscription-simulation.constants.ts
+++ b/client/app/cross-modules/utilities/subscription-simulation/constants/subscription-simulation.constants.ts
@@ -2,6 +2,8 @@ export const SUBSCRIPTIONS_ENDPOINT = "/api/subscriptions";
 export const SUBSCRIPTIONS_CURRENT_ENDPOINT = "/api/subscriptions/current";
 export const ENTITLEMENTS_ENDPOINT = "/api/entitlements";
 export const SUBSCRIPTION_USAGE_ENDPOINT = "/api/subscription-usage";
+export const SUBSCRIPTION_USAGE_OVERAGE_PREVIEW_ENDPOINT =
+  "/api/subscription-usage/overage/preview";
 
 /**
  * The test-harness controller, distinct from the integrator-facing API above: it forces payment
--- a/client/app/cross-modules/utilities/subscription-simulation/hooks/use-preview-usage-overage.ts
+++ b/client/app/cross-modules/utilities/subscription-simulation/hooks/use-preview-usage-overage.ts
@@ -1,0 +1,13 @@
+import { useMutation } from "@tanstack/react-query";
+import type { PreviewUsageOverageRequest } from "../models/subscription-simulation.model";
+import { subscriptionSimulationService } from "../services/subscription-simulation.service";
+
+/**
+ * What additional metered usage would cost. Writes nothing, so it invalidates nothing — the
+ * server computes every price; this hook only carries the request and hands back the response.
+ */
+export const usePreviewUsageOverage = () =>
+  useMutation({
+    mutationFn: (request: PreviewUsageOverageRequest) =>
+      subscriptionSimulationService.previewUsageOverage(request),
+  });
--- a/client/app/cross-modules/utilities/subscription-simulation/models/subscription-simulation.model.ts
+++ b/client/app/cross-modules/utilities/subscription-simulation/models/subscription-simulation.model.ts
@@ -359,3 +359,81 @@ export interface RecordUsageResult {
   overage: number;
   replayed: boolean;
 }
+
+/**
+ * A hypothetical slice of additional metered usage to price. Writes nothing and charges nothing —
+ * see {@link UsageOveragePreviewResult}.
+ */
+export interface PreviewUsageOverageRequest {
+  meterKey: string;
+  /** Cannot be zero or negative — a preview of no additional usage answers nothing. */
+  additionalQuantity: number;
+  /** Honoured only for this portal acting as the platform console; see SubscribeToPlanRequest. */
+  organizationId?: string;
+}
+
+/** One charge, split the way an invoice line would be. All amounts are minor units. */
+export interface UsageChargeAmounts {
+  grossMinor: number;
+  automaticDiscountMinor: number;
+  netMinor: number;
+  taxMinor: number;
+  totalMinor: number;
+}
+
+/** One graduated tier band's slice of the additional usage. */
+export interface UsageOverageTierAllocation {
+  /**
+   * The first overage unit this band covers, counted from the first overage unit of the whole
+   * period — not from wherever the additional usage itself begins.
+   */
+  fromOverageQuantity: number;
+  toOverageQuantity: number;
+  units: number;
+  unitAmountMinor: number;
+  amountMinor: number;
+}
+
+export interface UsageOveragePreviewDiscount {
+  automaticBasisPoints: number;
+  /** Always false — a promotional discount code never reduces metered overage. */
+  promotionalCodeApplied: boolean;
+}
+
+export interface UsageOveragePreviewTax {
+  rateBasisPoints: number | null;
+  mode: string;
+}
+
+/**
+ * What a hypothetical slice of additional metered usage would cost, rated with the subscription's
+ * own snapshotted terms and the same order of operations period-end usage rating uses — but
+ * nothing here is recorded and nothing here is charged. Server-computed throughout; nothing here
+ * should be recalculated in the browser.
+ */
+export interface UsageOveragePreviewResult {
+  meterKey: string;
+  unitLabel: string;
+  currencyCode: string;
+  periodKey: string;
+  periodStartUtc: string;
+  periodEndUtc: string;
+  /** When this preview was computed. Usage recorded afterward can change the eventual invoice. */
+  calculatedAtUtc: string;
+  includedQuantity: number;
+  currentUsage: number;
+  currentOverage: number;
+  additionalQuantity: number;
+  projectedUsage: number;
+  projectedOverage: number;
+  currentCharge: UsageChargeAmounts;
+  /** The difference between the projected and current charges — never rated on its own. */
+  additionalCharge: UsageChargeAmounts;
+  projectedPeriodCharge: UsageChargeAmounts;
+  additionalTierBreakdown: UsageOverageTierAllocation[];
+  discount: UsageOveragePreviewDiscount;
+  tax: UsageOveragePreviewTax;
+  writesUsage: boolean;
+  chargesPayment: boolean;
+  finalChargeDependsOnActualPeriodEndUsage: boolean;
+}
--- a/client/app/cross-modules/utilities/subscription-simulation/services/subscription-simulation.service.ts
+++ b/client/app/cross-modules/utilities/subscription-simulation/services/subscription-simulation.service.ts
@@ -4,6 +4,7 @@ import {
   AUDIT_TRAIL_DEFAULT_LIMIT,
   ENTITLEMENTS_ENDPOINT,
   SUBSCRIPTION_USAGE_ENDPOINT,
+  SUBSCRIPTION_USAGE_OVERAGE_PREVIEW_ENDPOINT,
   SUBSCRIPTIONS_CURRENT_ENDPOINT,
   SUBSCRIPTIONS_ENDPOINT,
 } from "../constants/subscription-simulation.constants";
@@ -15,6 +16,7 @@ import type {
   QuantityChangeQuote,
   EntitlementDecision,
   EntitlementsSnapshot,
+  PreviewUsageOverageRequest,
   RecordUsageRequest,
   RecordUsageResult,
   SimulatedSubscription,
@@ -22,6 +24,7 @@ import type {
   SubscriptionAuditEvent,
   SubscriptionPlanChangePreview,
   SubscriptionPurchasePreview,
+  UsageOveragePreviewResult,
 } from "../models/subscription-simulation.model";
 
 interface SimulationApiError {
@@ -427,6 +430,28 @@ class SubscriptionSimulationService {
 
     if (!response.success || !response.data) {
       throw new Error(response.error?.message || "The usage could not be recorded.");
+    }
+
+    return response.data;
+  }
+
+  /**
+   * Estimates the cost of additional metered usage against the caller's current subscription.
+   * Advisory only: the server neither records the usage nor charges anything, and every price in
+   * the response is server-computed — this is a thin fetch wrapper, not a place to recalculate.
+   */
+  async previewUsageOverage(
+    request: PreviewUsageOverageRequest,
+  ): Promise<UsageOveragePreviewResult> {
+    const response = await serviceInstances.utitlitiesService.post<
+      SimulationApiResponse<UsageOveragePreviewResult>
+    >(SUBSCRIPTION_USAGE_OVERAGE_PREVIEW_ENDPOINT, request);
+
+    if (!response.success || !response.data) {
+      throw new SubscriptionOperationError(
+        response.error?.message || "The overage could not be previewed.",
+        response.error?.code ?? "unknown",
+      );
     }
 
     return response.data;
--- a/server/Api/Api.xml
+++ b/server/Api/Api.xml
@@ -914,6 +914,17 @@
             must not exceed its allowance should set <c>enforce</c> and act on <c>allowed</c>.
             </remarks>
         </member>
+        <member name="M:Api.Controllers.SubscriptionUsageController.PreviewOverage(Subscription.DomainService.Requests.PreviewUsageOverageRequest,System.Threading.CancellationToken)">
+            <summary>
+            Estimates the cost of additional metered usage, using the active subscription's own
+            snapshotted terms and the same rating, discount and tax logic as the final usage invoice.
+            </summary>
+            <remarks>
+            Advisory only: nothing here is recorded against the usage ledger and nothing here is
+            charged. Usage recorded after <c>calculatedAtUtc</c> can change the eventual invoice — see
+            <c>finalChargeDependsOnActualPeriodEndUsage</c> on the response.
+            </remarks>
+        </member>
         <member name="T:Api.Controllers.TemplateEngineController">
             <summary>
             API Controller for managing template engine operations
--- a/server/Api/Controllers/SubscriptionUsageController.cs
+++ b/server/Api/Controllers/SubscriptionUsageController.cs
@@ -18,8 +18,15 @@ namespace Api.Controllers;
 public sealed class SubscriptionUsageController : ControllerBase
 {
     private readonly IUsageRecordingService _usage;
+    private readonly ISubscriptionUsageOveragePreviewService _overagePreview;
 
-    public SubscriptionUsageController(IUsageRecordingService usage) => _usage = usage;
+    public SubscriptionUsageController(
+        IUsageRecordingService usage,
+        ISubscriptionUsageOveragePreviewService overagePreview)
+    {
+        _usage = usage;
+        _overagePreview = overagePreview;
+    }
 
     /// <summary>
     /// Records usage and reports where that leaves the allowance.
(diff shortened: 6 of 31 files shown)


fahad0208:
Addressed all 5 findings in commit 5187bcd9.

**1. [P1] Carry-forward overage never invoiced** — `SubscriptionUsageRatingProcessor.EnsureInvoiceAsync` now iterates `subscription.Plan.Meters.Where(m => m.ResetPolicy != MeterResetPolicy.Never)` instead of `== Periodic`, so `CarryForward` meters are rated exactly like `Periodic` ones (per `MeterResetPolicy.CarryForward`'s own docs — it still resets and rates per window). Added `CarryForward_meters_are_rated_by_final_rating_not_silently_skipped`, which asserts the pre-fix filter would have produced zero invoice lines and the post-fix code produces a real one.

**2. [P1] Trial/carry-forward allowance disagreement** — `EnsureInvoiceAsync` now calls `IMeterAllowanceResolver.EffectiveAsync(subscription, meter, period, counter, ct)` (the same resolver the preview already used) to get the frozen/effective allowance, computes `overageQuantity = max(0, balance - allowance)`, and passes that quantity into `SubscriptionUsageRater.OverageAllocations` — never into `OverageAmountMinor`, which recomputed overage from bare `IncludedQuantity`. The processor now takes an optional `IMeterAllowanceResolver` constructor parameter (defaulting to a real instance when not supplied, consistent with its other optional collaborators), and `EnsureInvoiceAsync` takes a `BillingPeriod` instead of a bare `periodKey` string so the resolver has period bounds to work with; both call sites (the outgoing-plan-change/cancellation snapshot loop and the current-period sweep) now build one. Added `Trial_grant_allowance_agrees_between_preview_and_final_rating`, reproducing the reviewer's exact example (plan 150 / trial 500 / usage 300): it asserts what the old code would have billed (150 units via `OverageAmountMinor`, documented inline as the wrong answer) and that the fixed invoice actually charges 0, matching the preview.

**3. [P1] One-meter preview vs. aggregated invoice** — `SubscriptionUsageOveragePreviewService.PreviewAsync` now reads and rates every other billable meter (`ResetPolicy != Never`, excluding the requested one) at its own current allocation, sums that into `otherMetersAggregateGross`, then builds `currentAggregateGross`/`projectedAggregateGross` by substituting the requested meter's current/projected contribution into that aggregate. `UsageChargeCalculator.Charge` is applied once to each aggregate (mirroring how the worker discounts/taxes the whole invoice once), and `additionalCharge` is their difference — never rated in isolation. `currentCharge`/`projectedPeriodCharge` in the response now reflect these aggregates; `additionalTierBreakdown` stays scoped to the requested meter (informational only, as before). Added `Two_overage_meters_at_a_rounding_boundary_agree_with_the_aggregate_not_a_slice` with a 5% discount / 7.7% tax and two meters (screening 1,000 gross + storage 350 gross) chosen so the aggregate-correct answer (337) and the old isolated-slice answer (338) differ by exactly one minor unit — the test asserts the response lands on 337.

**4. [P2] Integration guide not updated** — Added a `POST /api/subscription-usage/overage/preview` section to `server/Api/Documentation/subscription/v1/index.html`, matching the existing style used for `/subscription-usage/current` and the other endpoints in that file: request table, a full sample response (mirroring the original spec's field list), prose on the aggregate `currentCharge`/`projectedPeriodCharge` semantics from fix #3, the `writesUsage`/`chargesPayment`/`finalChargeDependsOnActualPeriodEndUsage` flags, and the relevant `Returns` line. Also added the endpoint's named error codes (`subscription_usage_preview_invalid`, `subscription_meter_overage_not_allowed`, `subscription_meter_rate_unavailable`, `subscription_schedule_unavailable`) to the shared error-code table, and updated the domain README's own description of `currentCharge`/`projectedPeriodCharge` to describe the aggregate.

**5. [P2] Unchecked quantity addition** — `projectedUsage = currentUsage + request.AdditionalQuantity` is now inside a `checked` block; an `OverflowException` is caught and returns `subscription_usage_preview_invalid` (400, `PaymentFailureKind.Validation`) instead of silently wrapping. Added `An_additional_quantity_that_would_overflow_the_projection_is_a_named_failure`.

**Verification**
- `dotnet build server/Api/Api.csproj --no-incremental`: 0 errors.
- `dotnet test --filter "FullyQualifiedName!~Integration"` from `server/`: 3,107 passed, 1 skipped, 4 failed — all 4 are the pre-existing `SubscriptionSimulationGuardTests`/`SubscriptionSimulationDataConsoleServiceTests`/`SubscriptionSimulationServiceTests` permission-guard failures, unrelated to this change (confirmed via `git diff origin/dev..HEAD` showing zero diff in `Subscription.DomainService/Simulation/` or its test files).
- `npm run type-check` from `client/`: 0 errors — the preview response TS shape is unchanged; only the semantics of `currentCharge`/`projectedPeriodCharge` changed (single-meter → aggregate), not their fields.

Commit: 5187bcd9

fahad0208:
Pushed 6dd34b54, addressing the remaining P1 finding and the overflow hardening item.

## P1: cut-short periods now preserve the original allowance context

Added `PendingUsagePeriod.MeterAllowances` (`Dictionary<string, long>?`), snapshotted via the same `IMeterAllowanceResolver.EffectiveAsync` the live resolver and preview both use — captured at the exact call sites that build the outgoing period (`SubscriptionCancellationService.OutgoingUsagePeriodOfAsync`, `SubscriptionCancellationEffectiveProcessor.OutgoingUsagePeriodOfAsync`, `SubscriptionPlanChangeService.SnapshotOutgoingUsagePeriodAsync`), **before** the compare-and-set that changes status or re-anchors `UsageSchedule` is applied — same ordering discipline the existing `OutgoingUsagePeriod`/closure-reservation snapshots already follow. The capture logic is shared in a new `MeterAllowanceSnapshot.CaptureAsync` helper so all three call sites agree.

`SubscriptionUsageRatingProcessor.EnsureInvoiceAsync` now takes the frozen snapshot value per meter when present, and only falls back to a live resolve (against the synthetic rating subscription's current status/trial/schedule, as before) when the snapshot has no entry for that meter — which is exactly the **legacy-document case**: a `PendingUsagePeriod` written before this field existed has `MeterAllowances == null`, and every meter in it takes the fallback path, unchanged from prior behavior. Nothing about existing queued documents needs a migration or backfill; they keep rating exactly as they did before this change, they just don't get the fix's benefit retroactively.

Tests added (`SubscriptionCancellationServiceTests`, `SubscriptionPlanChangeServiceTests`, `SubscriptionUsageRatingProcessorTests`):
- `An_immediate_cancellation_mid_trial_freezes_the_trial_grant_not_the_post_cancellation_allowance` — cancels mid-trial and asserts the snapshot carries the trial grant (300), not the plan's plain allowance (500).
- `A_downgrade_freezes_the_carried_forward_allowance_before_the_schedule_re_anchors` — a real carry-forward scenario with `CarryForwardCap = 50` and a previous window at 80/100 used, asserting the snapshot captures 120 (100 base + 20 actually carried). Deliberately not the `CarryForwardCap = 0` shape the prior round's test used, which the plan validator would reject and which never exercises the carry math.
- `A_pending_periods_frozen_allowance_snapshot_is_used_even_when_the_counter_never_existed` — the actual crash scenario: ledger balance present, no counter at all, frozen snapshot of 300; asserts final rating uses 300 (4,000 overage) and not the live-resolved 500 (which would have under-billed to 2,000).
- `A_legacy_pending_period_with_no_allowance_snapshot_falls_back_to_the_live_resolver` — `MeterAllowances` left null, confirming the exact pre-existing behavior and figure (2,000) is preserved.

## Overflow hardening in tier/aggregate math

`SubscriptionUsageRater.WalkTierRange` now performs its `Int128` tier multiplication and every narrowing cast back down to `long` (per-tier `AmountMinor` and the final total) inside a `checked` block, throwing `OverflowException` instead of silently wrapping.

- `SubscriptionUsageOveragePreviewService` wraps the tier-allocation calls and the aggregate-gross additions (per-meter running sum and the current/projected aggregate sums) in `checked` and a `try/catch`, returning `subscription_usage_preview_invalid` on overflow — same named failure the existing `currentUsage + additionalQuantity` guard uses.
- `SubscriptionUsageRatingProcessor.EnsureInvoiceAsync` has no HTTP response to refuse with, so it catches the overflow at both the per-meter tier-allocation call and the final line-sum/charge call, logs an error, and returns without creating an invoice for that period — deferring it to a later sweep pass rather than ever persisting a wrapped total. This mirrors the existing pattern in that method for `BillingPeriodCalculator.TryGetPeriod` failing.

Tests added:
- `SubscriptionUsageRaterTests.A_tier_total_that_would_overflow_a_long_throws_rather_than_wraps` and the companion `A_large_but_valid_tier_total_still_prices_correctly` (proving the checked arithmetic doesn't reject charges that were never actually going to overflow).
- `SubscriptionUsageOveragePreviewServiceTests.A_tier_total_that_would_overflow_a_long_is_a_named_failure_not_a_wrapped_charge`.
- `SubscriptionUsageRatingProcessorTests.A_period_whose_overage_would_overflow_a_long_is_deferred_rather_than_mispriced`.

## Verification

- `dotnet build server/Api/Api.csproj --no-incremental` — 0 errors.
- `dotnet test --filter "FullyQualifiedName!~Integration"` from `server/` — 3120 tests, 3115 passed, 4 failed, 1 skipped. The 4 failures are the pre-existing, unrelated `SubscriptionSimulationServiceTests`/`SubscriptionSimulationDataConsoleServiceTests`/`SubscriptionSimulationGuardTests` permission-check failures already present on `origin/dev`, not introduced by this change.
- `npm run type-check` from `client/` — 0 errors (this round is server-only, confirmed as a no-op check).

Scope note: I did not add a full end-to-end integration test running the actual `SubscriptionCancellationService`/`SubscriptionPlanChangeService` output through `SubscriptionUsageRatingProcessor` in one test (i.e. capture via the real service, then rate via the real processor, in a single test method) — the trial and carry-forward tests verify the snapshot capture side, and the crash-path/legacy tests verify the rating side, using the same `IMeterAllowanceResolver` both layers already share. If you want the literal single-test round trip in addition to these, I can add it.

fahad0208:
Pushed a fix for both P1 findings from this round, commit b8ca415a47b95fa2cb15015c0f3717fa1cf445b2.

**Finding 1 — deferred periods were actually discarded**

`EnsureInvoiceAsync` (`server/Subscription.DomainService/Outbox/SubscriptionUsageRatingProcessor.cs`) now returns an explicit `InvoiceReadiness { Ready, Deferred }` instead of `Task` — following the same "outcome enum" convention as `ClosureCommitOutcome`/`ClosureReservationOutcome` in this file's own dependencies. `Ready` covers both "invoice already existed" and "created now"; `Deferred` covers both existing overflow catches (tier-walk overflow, and now the tax overflow described below) — each already logs an error, unchanged.

Both call sites now branch on the outcome:
- Pending-period path (cut-short window from cancellation/plan-change): on `Deferred`, `continue`s immediately — does **not** schedule the invoice charge, does **not** call `TryRemovePendingUsagePeriodAsync`, does **not** call `TryMarkClosedAsync`. The `PendingUsagePeriod` snapshot and the closure's `Closing` state are left exactly as they were.
- Live-period path: on `Deferred`, `break`s out of the sweep loop for that subscription instead of calling `TryTransitionAsync` — the usage clock is not advanced past the unbilled period.

**How I verified this is genuinely retried, not just "not marked done":** `CloseSubscriptionAsync` is invoked per subscription every time `CloseDuePeriodsAsync` runs its sweep, and `ListDueForUsageRatingAsync` is what decides which subscriptions are "due" — that query is driven by `CurrentUsagePeriodEndUtc <= now` for the live path and by the mere presence of an unprocessed `PendingUsagePeriods` entry for the cut-short path. Since a deferred outcome leaves `CurrentUsagePeriodEndUtc` unmoved and the `PendingUsagePeriods` entry in place, the subscription is trivially still "due" on the next call — no new retry queue or scheduling was needed, exactly as the existing comments in this file already describe for other not-yet-ready cases (an outstanding claim, a schedule whose time zone briefly failed to resolve). I exercised this directly in tests rather than just reasoning about it: `A_period_whose_overage_would_overflow_a_long_is_deferred_rather_than_mispriced` now also asserts `TryTransitionAsync` is never called and the sweep's own `closedCount` return is `0`; a new `A_pending_periods_overflow_leaves_its_snapshot_and_closure_untouched` test asserts `TryRemovePendingUsagePeriodAsync` and `TryMarkClosedAsync` are never called for the deferred period.

**Finding 2 — tax-exclusive totals could still overflow silently**

`SubscriptionAmountCalculator.TaxBreakdownFor`'s exclusive-tax branch now wraps `discountedAmountMinor + exclusiveTax` in a `checked` block. While in there I also found `RoundedQuotient`'s own `Int128`→`long` narrowing cast was unchecked — `checked`/`unchecked` context does not cross a method call, so wrapping the caller alone would not have covered it; `RoundedQuotient` now checks its own cast. Both the null-taxMode legacy path and the explicit-mode path route through this.

This throws `OverflowException` from deep inside the shared calculator, which both consumers already had (or now have) a catch for:
- `SubscriptionUsageRatingProcessor.EnsureInvoiceAsync`'s existing try/catch around `UsageChargeCalculator.Charge(...)` already covered this — it now returns `InvoiceReadiness.Deferred` through the exact same mechanism as Finding 1, so a tax overflow during final rating is preserved/retried, not silently skipped-but-marked-done. This composes correctly without any special-casing.
- `SubscriptionUsageOveragePreviewService.PreviewAsync` did **not** have a guard around its `UsageChargeCalculator.Charge`/`Difference` calls — that was a real gap; an overflow there would have bubbled up as an unhandled 500. Added a try/catch translating `OverflowException` into the existing `subscription_usage_preview_invalid` failure, the same code the tier-walk overflow guard already used.

Added tests for both consumers using a gross overage that comfortably fits a `long` on its own (single overage unit at a very large unit rate) but overflows once tax is layered on top — a distinct failure point from the existing tier-walk overflow tests, which never reach the tax step at all.

**Scoped down:** I did not touch `BuiltInDiscountCalculator`'s own `Int128`→`long` narrowing cast (used for the automatic-discount/volume-band math) — it wasn't flagged by this round's findings, is a subtraction-dominated path less exposed to this same failure mode, and expanding scope there felt like a separate change. Flagging it here in case it's worth a follow-up.

**Build/test/type-check:**
- `dotnet build server/Api/Api.csproj --no-incremental` — 0 errors (3688 warnings, all pre-existing analyzer noise unrelated to this change).
- `dotnet test --filter "FullyQualifiedName!~Integration"` from `server/` — 3118 passed, 4 failed, 1 skipped. The 4 failures are `SubscriptionSimulationGuardTests`/`SubscriptionSimulationServiceTests`/`SubscriptionSimulationDataConsoleServiceTests` permission-check failures; confirmed pre-existing by diffing those test files against `origin/dev` (no diff — untouched by this branch).
- `npm run type-check` from `client/` — 0 errors (this round is server-only; confirmed as a no-op check).

PR #340, opened 2026-08-28
ci: compile and test the client on pull requests

fahad0208:
## Why

Nothing validated the client on a pull request. The diagnosis is not a path filter — `build-client` in `ci-dev.yml` is gated on:

```yaml
if: |
  github.event_name == 'push' &&
```

It exists to publish an image from `dev`, not to check a change before it lands. The `pr-checks` job beside it is commented out and was `test-dotnet.yml` in any case, and `RUN_TESTS` is `"false"`.

So a client-only pull request showed an all-green status having never had its TypeScript compiled or its ~2,240 tests run. The ticks came from CodeQL, `repo-hygiene` and `initialization`. Three client PRs merged on that basis this week.

## What gates

`.github/workflows/client-pr-validation.yml`, on pull requests to `dev`/`main` touching `client/**`, and on pushes to those branches so a bad merge is attributed to the merge:

| Step | Gates | On `dev` today |
|---|---|---|
| `npm run type-check` | yes | exit 0 |
| `npm test` | yes | exit 0 *after this PR* |
| `npm run build` | yes | exit 0 |
| `npm run lint` | **no — advisory** | exit 1: 32 errors, 75 warnings |

Steps use `if: always()` so one run tells you everything that is wrong, not just the first thing.

Type-check and build are both present deliberately: `build` runs `tsc -b`, which honours project references and can skip a project whose inputs look unchanged, while `type-check` is `tsc --noEmit` over the whole app config. The build also catches what only bundling finds.

**Lint is advisory because it does not pass on `dev`.** 32 errors, none introduced here. Gating on it would make this job red on arrival, and a job that is red on arrival gets switched off within the week. The comment says to flip `continue-on-error` off once the count reaches zero — that is a decision about 32 unrelated files, not one for this PR.

## Three things the gate had to be given a chance to pass

Each existed before today and was invisible *because* nothing ran the client on a PR.

**Four suites never ran at all.** `github-info.ts` imported five SVGs out of `client/public`, a directory Vite serves verbatim and deliberately does not process. The browser tolerated it; under Vitest the same specifier reaches Node as `file:///assets/github-icon.svg` and throws before collection, so `create-project`, `repository-selection-modal`, `create-project-resources-form` and `render-provider` failed to load. They are plain paths now — the same value the import produced, and what a sibling call site already wrote. **22 tests run for the first time.**

**Three tests were permanently red.** `step-trial.test.tsx` covered the "Require a payment method before activation" checkbox, which moved to the pricing step. The feature is intact; the tests were pointing at the wrong component. Moved to a new `step-pricing-model.test.tsx`, with the trial's own card question left behind in `step-trial.test.tsx` and each file now asserting the other's control is *not* present — the two read almost identically and mean different things.

**The 5s test timeout does not fit this suite.** `userEvent` types one keystroke at a time, so a test filling four fields spends seconds of wall clock waiting. Six unrelated files failed on timeout under load and passed when run alone; `create-payment-provider-page` needs **17s of test time on an idle machine**. A shared CI runner is a busy machine by definition. Raised to 20s in `vitest.config.ts`.

That last one is a global change where the narrower fix is `userEvent.setup({ delay: null })` in the slow tests. I took the one-line option rather than editing several dozen files, and said so in the config. It delays reporting a genuinely hung test; it does not hide one.

## Verification

- `npm run type-check` — exit 0. `npm run build` — exit 0.
- Relocated tests: 6 passed across the two step files.
- Previously-uncollectable suites: 4 files, 22 passed.
- Full suite before the timeout raise: 2238 passed, 1 failed (`create-payment-provider-page`, timeout). A confirming full run on `5b363c70` is in progress and I will post the count here either way.

Not addressed, and worth deciding separately: whether this becomes a required check. If it does, note that the `paths` filter means it does not run on server-only PRs, which GitHub can treat as a missing required check.


fahad0208:
### Confirming run on `5b363c70`: green

```
Test Files  319 passed (319)
     Tests  2239 passed (2239)
```

`npm test` exits 0. No failures, no unhandled errors, nothing skipped by a load error.

The same command on this branch before the timeout raise: **2238 passed, 1 failed** — `create-payment-provider-page`, timed out at 5s. That was the last one, and it is the reason the raise belongs in this PR rather than in a follow-up: without it the gate this PR installs would have been red on its own first run.

For comparison against `dev`, where the gate does not exist:

| | `dev` | this branch |
|---|---|---|
| Suites that load | 315 | **319** |
| Tests that run | 2217 | **2239** |
| Failures | 3 (`step-trial`) + 4 suites unloadable | **0** |

The 22-test difference is the four suites that could never be collected. They were not failing — they were absent, which is worse, because absent coverage reads as no news.

`type-check` and `build` both exit 0. Lint is unchanged and still advisory: 32 errors, 75 warnings, all pre-existing.

PR #357, opened 2026-08-29
feat(subscription): reconciliation sweep for stuck campaign redemptions [Phase 6/6]

fahad0208:
Phase 6/6 — the final phase of campaign discount management. Closes the crash window flagged since Phase 2: a subscription's own transition (activation, expiry, cancellation) committing while the paired campaign ledger call never runs, or runs only halfway.

## What was already closed, and what was not

`TryReleaseAsync` already narrows half of this itself — it commits a durable `ReleasePending` state before its second step, specifically so "a reconciliation sweep can find and finish" a crash between the two (see that method's own remarks, written in Phase 2). What Phase 2 left open, by its own admission:

1. A `Reserved` row whose subscription actually activated, where the crash landed **before `TryMarkRedeemedAsync` ever ran at all** — there's no intermediate state for the redeem path the way there is for release, so this row looks identical to one nothing has happened to yet.
2. A row already at `ReleasePending` whose second step never ran.

Both leave a campaign redemption stuck forever — for a one-use campaign, the organization's slot never frees, indistinguishable from "still in use" even though the subscription's fate was already decided.

## `CampaignRedemptionReconciler` (new)

Reads a subscription's `ActivatedAtUtc` rather than its current status: that field is set once, at activation, and never cleared afterward — including by a later cancellation — so it answers "did this subscription's campaign ever get spent" independent of whatever status the subscription holds by the time the sweep runs.

- `ActivatedAtUtc` set → `TryMarkRedeemedAsync`, backdated to that instant.
- Terminal and never activated (`IncompleteExpired` or `Canceled`-while-`Incomplete`) → `TryReleaseAsync`, which resumes correctly whether the row is still at `Reserved` or already at `ReleasePending`.
- Still `Incomplete` → left alone; its own `ActivationRecovery` work resolves that first.

Wired into `SubscriptionRepairAnnouncer.AnnounceAsync` directly, not through the durable work queue — like `ReconcileStaleClosuresAsync` beside it, this moves no money, so it needs none of the two-executor discipline that queue exists to protect. Threaded through as an optional constructor dependency, the same convention every other campaign collaborator in this feature follows.

## Repository additions

- `ListStaleAsync(tenantId, reservedBeforeUtc, limit)` — `Reserved` or `ReleasePending` rows, oldest first. Backed by a new, deliberately **non-partial** index: `State`'s two reconciled values (0 and 2) straddle `Redeemed` (1), so no ordered comparison in a partial filter *definition* could isolate them — but an ordinary compound index serves an `$in` query over `State` fine, since that restriction is specific to how a partial filter is defined, not to how a query is written.
- `SubscriptionOptions.CampaignRedemptionGraceMinutes` (15) / `CampaignRedemptionBatchSize` (50) — same shape as `SettlementReservationGraceMinutes`/`BatchSize` beside them.

## Tests

- `CampaignRedemptionReconcilerTests` (8 tests, mocked): every branch above, plus a redemption naming a subscription that can't be found being left alone rather than guessed at, and the grace period actually coming from options.
- `CampaignRedemptionConcurrencyTests` (+2, real MongoDB): `ListStaleAsync` finds only `Reserved`/`ReleasePending` rows before the threshold — proving the `$in`-over-a-non-partial-index reasoning against a real server rather than arguing it in a comment — ordered oldest first, respecting the limit.

## Verification

Server: `dotnet test --filter "FullyQualifiedName!~Integration"`: **3225 passed**, same 4 pre-existing baseline failures, 1 unrelated skip. `CampaignRedemptionConcurrencyTests`: 9/9 passed in isolation against a fresh local MongoDB.

One thing flagged rather than fixed: running the **entire** integration suite together against one container reproducibly fails 16 unrelated tests across three other classes, on two separate completely fresh containers — none campaign-related, none touched by this change. Pre-existing full-suite fragility, not a regression from this phase.

This is the last of the six phases from the original plan.

## Merge order

Based on `feat/campaign-admin-authoring-ui` (#356) → `feat/campaign-buyer-preview` (#354) → `feat/campaign-annual-yearly-pricing` (#352) → `feat/campaign-free-month-pricing` (#351) → `feat/campaign-discounts-redemption-ledger` (#350) → `feat/campaign-discounts-data-model` (#349). Merge in that order, retargeting each PR's base as the prior one lands.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/client/app/cross-modules/utilities/subscription/components/campaign-builder/campaign-builder.tsx
+++ b/client/app/cross-modules/utilities/subscription/components/campaign-builder/campaign-builder.tsx
@@ -34,6 +34,8 @@ export interface CampaignBuilderProps {
   /** Rejecting leaves the draft in place and shows the error; the caller navigates on success. */
   onSubmit: (draft: CampaignDraft, organizationId: string | undefined) => Promise<void>;
   onCancel: () => void;
+  initialDraft?: CampaignDraft;
+  editing?: boolean;
 }
 
 export const CampaignBuilder = (props: CampaignBuilderProps) => (
@@ -49,9 +51,11 @@ const CampaignBuilderWizard = ({
   submissionError,
   onSubmit,
   onCancel,
+  initialDraft = EMPTY_DRAFT,
+  editing = false,
 }: CampaignBuilderProps) => {
   const { currentStep, nextStep, previousStep } = useStepper();
-  const [draft, setDraft] = useState<CampaignDraft>(EMPTY_DRAFT);
+  const [draft, setDraft] = useState<CampaignDraft>(initialDraft);
   const step = currentStep as StepId;
   const isLastStep = currentStep === STEPS.length;
 
@@ -82,7 +86,7 @@ const CampaignBuilderWizard = ({
       <CampaignBuilderProgress />
 
       <Card className="space-y-5 rounded-2xl p-5 sm:p-7">
-        {step === 1 && <StepIdentity draft={draft} onChange={update} />}
+        {step === 1 && <StepIdentity draft={draft} onChange={update} codeReadOnly={editing} />}
         {step === 2 && <StepBenefit draft={draft} onChange={update} />}
         {step === 3 && <StepEligibility draft={draft} plans={plans} onChange={update} />}
         {step === 4 && <StepReview draft={draft} plans={plans} />}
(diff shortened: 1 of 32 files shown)


fahad0208:
Implemented the review-gap fixes in commits f409c424 and 00553d64. Highlights: campaign-level payment-method requirements now drive zero-amount checkout; FirstAnnualPeriod discounts both opening stub and first year without spending the year on the stub; free-opening campaigns can be open-ended; buyer-safe POST /api/subscription-discounts/preview returns a normal quote plus Applied/NotFound/NotStarted/Expired/NotApplicable/AlreadyRedeemed/Unavailable without writes; free-period plan/quantity previews expose blockers; reconciliation defers unresolved rows to prevent page starvation; admin UI supports optimistic-version edits and shows redemption counts; catalogue mutations are audited; guide/README updated. Verification: API and XUnit projects build with 0 errors; 59 focused server tests pass; full non-integration 3225 passed with the 4 known simulation failures plus one known timing flake; client type-check and production build pass; 23 campaign-draft tests pass; diff check clean.

PR #381, opened 2026-08-31
feat(subscription): irreversible plan archiving and a catalogue built to manage it

fahad0208:
Archiving takes a whole plan off the menu permanently. Everyone already subscribed carries on untouched — they bill from the terms copied onto their subscription when they bought, so renewals, usage rating, entitlements, invoicing and cancellation all continue. None of the plan's prices is rewritten or deleted. What stops is selling.

## What was already true, and left alone

Worth stating first, because the spec reads as though this were a blank slate and it was not:

| Already present | Where |
| --- | --- |
| `CatalogueStatus { Draft, Active, Archived }`, `Plan.Status`, `Version`, `LastUpdatedDateUtc` | **no schema migration needed** |
| Subscriber-facing lists already excluded archived plans | `ListPlansAsync` filtered `Active` |
| Creation, purchase preview, plan-change preview and execution already refused them | all four resolve through `FindPlanByCodeAsync`, which filtered `Active` |
| Duplication | already worked via router state on the create page — the card's menu reuses it rather than adding a second mechanism |

So the enforcement largely existed; what it lacked was a **name**. `subscription_plan_archived` is now returned by consulting the archived catalogue *only after* the active lookup has found nothing.

**That ordering is load-bearing, not an implementation detail.** Resolving both statuses together would let an organization's archived plan shadow the tenant's active plan of the same code and refuse a sale that should have gone through — a revenue bug invisible to any test using one scope. There is a test for exactly this.

## Server

`PUT /subscription-plans/{planId}/archive` — conditional on status **and** on the version just read. `ModifiedCount == 0` covers three situations needing three answers, so the service re-reads to tell them apart:

| What happened | Answer |
| --- | --- |
| Already archived | success, no second write |
| Another archive won the race | success — both wanted this state |
| An unrelated edit moved the version on | `subscription_plan_changed` |
| Gone, or invisible to the caller | ordinary not-found |

Idempotency is the one place this differs from `ArchivePriceAsync`, which reports a repeat as a conflict. Commented, or the next reader will "fix" it.

**Five catalogue mutations refuse an archived plan** through one guard — update plan, create price, price tax, price discount, retire price. All five exist only to change what a *future* subscriber pays. Reading stays open: inspecting an archived plan is what the `Archived` filter is for.

`?status=Active|Archived|All`, invalid → `400 subscription_plan_status_invalid`. `Active` keeps the organization-over-tenant collapse; the archived views deliberately do not collapse by code.

Audit events gain nullable `AggregateType`/`AggregateId`/`AggregateCode` plus a sparse index. Every outcome is recorded, including the ones that changed nothing, and the audit write can never fail the archive it describes.

## Client

The catalogue fetches `All` once and filters in the browser, so tab switching is instant and the summary counts stay consistent with what the tabs show.

`useSubscriptionPlans` takes an optional status defaulting to `Active`, and **only this page passes anything**. The discount page and the subscribe/change-plan dialogs omit it and therefore cannot show an archived plan even by accident — a stronger guarantee than three screens each filtering for themselves.

Status, search and sort live in the query string and nowhere else, so a reload and a shared link reproduce the same list.

Some smaller decisions with reasons:

- **The card stopped being one big link.** A menu inside an anchor either navigates when it opens or fights the anchor to stop it.
- **Archived plans lose Edit and Archive entirely, not disabled.** The server refuses every mutation on one, so a greyed-out button invites a click and then an explanation. There is no restore to offer.
- **Archived cards are tinted, not faded.** Text at 60% opacity on a muted panel fails contrast, and an archived plan is still read.
- **A plan with no `status` reads as active everywhere.** Absent means a response cached before the field existed; the alternative mutes the whole catalogue after a deploy.
- **Price sorting compares major units.** On raw minor units `500 JPY` outranks `4.00 CHF` — not a conversion problem, just a different-sized smallest coin. Priceless and undated plans sort last, not first.
- **Double submission is guarded in the handler**, not only by `disabled`: a second Enter arrives before React re-renders, and this cannot be undone.

## Decisions where I followed your spec over my own first instinct

**Draft is not archivable** and answers as not found. I had originally argued for allowing it so a legacy draft could not be stranded unarchivable; your revision was explicit, and a draft appears in no catalogue view, so there is no menu to take it off.

## Verification

- **Server: 3287 unit tests pass**, 13 new. The **4 failures are pre-existing on `dev`** — `SubscriptionSimulationGuard.IsAuthorized` has its permission check commented out and replaced with `return true` (commit `8d7a4389`). Unrelated to this branch, and still unresolved.
- **Client: 2375 tests pass** (328 files), 26 new. `npm run type-check` and `npm run build` both exit 0.

## Not done

- **No page-level catalogue test.** The filter logic is covered directly and the card and dialog are covered as components, but the wiring between the URL and the rendered page is not asserted end-to-end.
- Responsive and keyboard behaviour is implemented (`sm`/`lg` grids, stacked filters, focus-visible rings, `aria-label`s on every control, `role="alert"`/`role="status"`) but **verified by reading, not by a test or a real browser**.

### DIFF
--- a/.github/workflows/subscription-scheduling-integration.yml
+++ b/.github/workflows/subscription-scheduling-integration.yml
@@ -48,12 +48,22 @@ jobs:
       # code. Left out, the suite reported 48 passing tests on the pull request that introduced it
       # and had tried none of them.
       #
+      # PlanArchiveIntegrationTests is named for the same reason: archiving is a status check and a
+      # version check in one conditional write, so "two writers race and exactly one wins" and
+      # "an archived organization plan does not hide the active tenant plan" are claims about
+      # MongoDB rather than about our code.
+      #
+      # SubscriptionCatalogueRepositoryIntegrationTests joins it here having run nowhere since it
+      # was written. It exists because TryUpdatePlanAsync builds its update from an explicit field
+      # list, so a field the service sets and the repository forgets is silently never persisted —
+      # which is exactly the kind of thing a filter that omits the class cannot report.
+      #
       # Several integration classes are still outside this filter and so run nowhere. Widening it
-      # belongs in its own change, since whatever it turns up would arrive as unrelated failures on
-      # whichever pull request happened to widen it.
+      # further belongs in its own change, since whatever it turns up would arrive as unrelated
+      # failures on whichever pull request happened to widen it.
       - name: Run subscription Mongo integration suite
         run: >-
           dotnet test server/XUnitTest/XUnitTest.csproj
           --no-restore
-          --filter "FullyQualifiedName~SubscriptionWorkQueueIntegrationTests|FullyQualifiedName~SubscriptionRenewalCrashRecoveryIntegrationTests|FullyQualifiedName~SubscriptionRepositoryIntegrationTests|FullyQualifiedName~SubscriptionQueueDocumentFlowIntegrationTests|FullyQualifiedName~CampaignRedemptionConcurrencyTests"
+          --filter "FullyQualifiedName~SubscriptionWorkQueueIntegrationTests|FullyQualifiedName~SubscriptionRenewalCrashRecoveryIntegrationTests|FullyQualifiedName~SubscriptionRepositoryIntegrationTests|FullyQualifiedName~SubscriptionQueueDocumentFlowIntegrationTests|FullyQualifiedName~CampaignRedemptionConcurrencyTests|FullyQualifiedName~PlanArchiveIntegrationTests|FullyQualifiedName~SubscriptionCatalogueRepositoryIntegrationTests"
           --verbosity minimal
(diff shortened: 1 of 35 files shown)


fahad0208:
Fixed the two remaining Mongo gaps in commit \10760881\:\n\n- The integration test now reads the real tenant database's \SubscriptionAuditEvents\ collection (the preceding \4bb1787\ correction), rather than a fabricated tenant-prefixed collection name.\n- The aggregate audit partial index now requires \AggregateType\ and \AggregateId\ to be BSON strings. \$exists\ was insufficient because ordinary nullable properties serialize as explicit BSON nulls.\n- The integration test now inserts both an ordinary subscription audit event and a plan aggregate event, then reads through the partial index and proves that only the aggregate event is a member. It no longer merely checks plausible-looking index metadata.\n\nThe server/test project compiles cleanly. Local execution reached the expected no-mongod limitation; CI's Mongo service will execute the strengthened test.

PR #386, opened 2026-08-31
feat(subscription): show complete tax breakdown on subscription purchase preview

fahad0208:
## Summary

Extends `POST /api/subscriptions/preview` so a client can explain both the tax due now and the tax included in the first renewal. All existing response fields and financial calculations are preserved unchanged; every addition is additive.

## API changes

New, additive fields on `SubscriptionPreviewResponse`:

```json
{
  "netSubtotalMinor": 2806,
  "tax": { "rateBasisPoints": 810, "mode": "Exclusive", "amountMinor": 227 },
  "nextRenewal": {
    "subtotalMinor": 14500,
    "builtInDiscountMinor": 0,
    "promotionalDiscountMinor": 0,
    "discountMinor": 0,
    "netSubtotalMinor": 14500,
    "tax": { "rateBasisPoints": 810, "mode": "Exclusive", "amountMinor": 1175 },
    "totalMinor": 15675,
    "renewalAtUtc": "2026-08-31T22:00:00Z"
  }
}
```

- `tax` is `null` **only** when the price carries no tax configuration at all — a card-free trial due nothing today, on a taxed price, reports `{ rateBasisPoints, mode, amountMinor: 0 }`, never `null`.
- `tax`/`nextRenewal.tax` are read from the subscription's snapshotted `PriceSnapshot.TaxRateBasisPoints`/`TaxMode` (via the already-shared `SubscriptionTaxPresentation.Describe`), never a second, live catalogue read — proven by a test asserting `GetPriceAsync` is called exactly once per preview.
- `nextRenewal` is built from the exact same `PeriodCharge` that the legacy `NextRenewalAmountMinor` already reads (captured once, reused for both) — `nextRenewal.totalMinor` always equals `NextRenewalAmountMinor`, `nextRenewal.renewalAtUtc` always equals `NextRenewalAtUtc`. No second tax/renewal calculation path exists anywhere in this change.
- `netSubtotalMinor` is `subtotalMinor` less every discount, before tax — the always-true identity a client can rely on is `netSubtotalMinor + (tax?.amountMinor ?? 0) === totalDueNowMinor`, in both tax modes.

## Client (subscription-simulation QA console)

The subscribe preview panel now shows an explicit **Subtotal → Built-in discount → Promotional discount → Net subtotal → Tax → Total** breakdown for both the due-now figures and the next-renewal figures (previously a single "First/Next renewal" line with only a total). Tax renders as `VAT (8.1%, included)` or `VAT (8.1%, added)` whenever `tax` is present, including a configured-but-zero amount — hidden only when `tax` is `null`.

## Tests

- **Server** (`SubscriptionCreationServiceTests.cs`, `CalendarAnnualChargeTimingTests.cs`): exclusive tax added on top of net subtotal; inclusive tax extracted from the discounted total; untaxed price returns `tax: null` on both breakdowns; card-free trial reports zero due-now tax with configured metadata and non-zero renewal tax; a discounted renewal's tax/total match the shared `PeriodCharge`; adding tax to a prorated calendar-aligned annual purchase preserves the existing due-now aggregation exactly, only adding the new tax figures; the price is read from the catalogue exactly once per preview.
- **Client** (`subscribe-dialog.test.tsx`): exclusive/inclusive rendering, configured-zero-tax during a trial, tax hidden only when null, due-now vs. renewal breakdowns rendering separately, and that discount/proration/campaign/total displays are unchanged.
- Both suites verified by temporarily breaking the change (nulling the `Tax` assignments server-side; hiding the tax row on a zero amount client-side) and confirming exactly the dependent tests fail while everything else still passes, then restoring the fix.

## Docs

`server/Api/Documentation/subscription/v1/index.html`'s preview section gains the new fields in its sample response, an inclusive and an exclusive example, and the `tax: null` vs. `amountMinor: 0` distinction spelled out explicitly.

## Verification

- `dotnet test server/XUnitTest` (excl. Integration): 3311 passed, 4 failed — the same pre-existing `SubscriptionSimulation*` permission-guard baseline failures already on `dev`, unrelated to this change.
- `npx vitest run` (client, full suite): 332 files / 2407 tests, all passed.
- `npx tsc --noEmit` and `eslint` over the touched module: clean.
- `npm run build` (production): succeeds.

### DIFF
--- a/client/app/cross-modules/utilities/subscription-simulation/components/subscribe-dialog.test.tsx
+++ b/client/app/cross-modules/utilities/subscription-simulation/components/subscribe-dialog.test.tsx
@@ -486,7 +486,7 @@ describe("SubscribeDialog", () => {
         netSubtotalMinor: 8_900,
         tax: null,
         totalMinor: 8_900,
-        renewalAtUtc: "2026-09-25T00:00:00Z",
+        renewalAtUtc: "2026-10-01T00:00:00Z",
       },
       // The actual next charge: a 6/30-day stub, genuinely cheaper than the recurring price above.
       nextCharge: {
--- a/client/app/cross-modules/utilities/subscription-simulation/components/subscribe-dialog.test.tsx
+++ b/client/app/cross-modules/utilities/subscription-simulation/components/subscribe-dialog.test.tsx
@@ -76,6 +76,21 @@ const quote: SubscriptionPurchasePreview = {
     totalMinor: 8_900,
     renewalAtUtc: "2026-09-16T00:00:00Z",
   },
+  nextCharge: {
+    chargeAtUtc: "2026-09-16T00:00:00Z",
+    periodStartUtc: "2026-09-16T00:00:00Z",
+    periodEndUtc: "2026-10-16T00:00:00Z",
+    prorated: false,
+    coveredDays: null,
+    totalDays: null,
+    subtotalMinor: 8_900,
+    builtInDiscountMinor: 0,
+    promotionalDiscountMinor: 0,
+    discountMinor: 0,
+    netSubtotalMinor: 8_900,
+    tax: null,
+    totalMinor: 8_900,
+  },
   trialEndsAtUtc: null,
   requiresCardSetup: false,
   pendingAnnualPeriod: null,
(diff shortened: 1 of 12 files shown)


fahad0208:
Fixed all three findings in commit \da7e09ae\.

- \
extCharge\ now honors \PendingAnnualPeriod\: boundary-collected annual charges use the frozen pending amount, while annual amounts already collected at checkout are not falsely shown as due again at the boundary.
- Ordinary future charges are priced at their actual charge instant and with projected opening-period discount consumption. Preview and activation now share \SubscriptionDiscountPeriodAccounting\, preventing their rules from drifting.
- \
extRenewal.renewalAtUtc\ now identifies the first full recurring charge. A prorated post-trial stub remains represented by \
extCharge\; the legacy \
extRenewalAtUtc\ field remains unchanged for compatibility.
- Updated contracts, client fixture, and published guide to explain the distinction.

Regression coverage includes both annual collection modes, a discount expiring before renewal, a one-period discount consumed by a calendar opening period, anniversary compatibility, and the corrected recurring date.

Verification: API build clean; 104 focused creation/activation/calendar tests pass; full non-integration suite previously completed with 3317 passing and only the same 4 pre-existing simulation-permission failures; subscribe-dialog tests 17/17; client type-check clean; \git diff --check\ clean.

PR #393, opened 2026-09-01
Merchant-selected Stripe/Adyen provider routing for subscriptions

fahad0208:
## Summary

Lets the platform console pick Stripe or Adyen per tenant on the merchant profile. New subscriptions route through whichever provider was selected at creation time; existing subscriptions keep whatever provider they were created with, since `BillingAccount.ProviderName` is frozen at creation and never re-read from the merchant profile by any later money operation.

## Scoping finding this PR confirms

Most of the provider-neutral plumbing already existed before this change — `BillingAccount.ProviderName` was already what `SubscriptionBillingGatewayResolver`, `RecurringChargeBillingGateway`, renewal/plan-change/quantity-change all keyed off, and Adyen already had config/secrets storage, hosted-checkout initiation, checkout status mapping, webhook normalization/verification for both webhook kinds, and the generic "Checkout API" gateway family (`CheckoutApiStoredPaymentChargeProviderGateway` etc.) already registered and resolved per-provider for refund/capture/stored-payment-charge. **This PR did not rebuild any of that.**

### Newly built
- `ISubscriptionPaymentProviderReadinessService` (`Subscription.DomainService/Services/`) — evaluates `Ready / Unsupported / NotConfigured / Disabled / Misconfigured / CredentialsUnavailable` for a `(tenantId, organizationId, providerName)`, reading `IPaymentRepository.GetProvidersAsync` (not the enabled-filtering `GetProviderAsync`, so `NotConfigured` and `Disabled` stay distinguishable) plus `IPaymentProviderSecretHydrator`. Backs the merchant-profile GET, its save validation, and subscription creation's pre-persist gate.
- `AdyenSetupSessionRequestFactory` (`Payment.DomainService/Providers/Adyen/`) — the zero-amount card-on-file setup session Adyen had no implementation for at all. Mirrors `AdyenInitiationRequestFactory`'s request shape (`amount.value: 0`, `storePaymentMethodMode: askForConsent`, `recurringProcessingModel: CardOnFile`, `shopperReference` always sent), registered alongside Stripe's `IPaymentMethodSetupRequestFactory`.
- `SubscriptionMerchantProfile.PaymentProviderName` (nullable — a legacy document with no value reads back as Stripe, computed only in `SubscriptionMerchantProfileService`, never defaulted at the entity/repository level), `UpdateMerchantProfileRequest.PaymentProviderName` (required, validated against the two supported providers), response fields `paymentProviderName` / `paymentProviderStatus` / `paymentProviders[]` (readiness for both providers, for the console's two selection cards), and audit logging on provider change (`ISubscriptionAuditTrail`, mirroring `PlanCatalogueService.RecordArchiveAsync`'s best-effort try/catch shape).
- Removed the two hardcoded `PaymentConstants.StripeProvider` call sites: `SubscriptionCreationService.BuildAsync` now resolves the merchant profile's selected provider (both the real-creation and write-free preview branches) and refuses to persist before any write if readiness isn't `Ready`; `SubscriptionCheckoutService`'s `CreatePaymentMethodSetupRequest`/`MakePaymentRequest` now read the subscription's already-frozen billing-account provider instead.
- `SubscriptionResponse.ProviderName` (server + mapper), wired at subscription creation/checkout's own response paths; **not wired at every `ToResponse` call site** (plan-change, cancellation, resume) — left `null` there. This is a scope cut for time, not a design decision; those callers would need a billing-account lookup added.
- Merchant-profile page UI: two provider-selection cards showing live readiness, Save disabled unless the selected provider is `Ready`, a link to the Payment Providers page when configuration is missing, and the confirmation copy from the spec ("This provider will be used for new subscriptions. Existing subscriptions will continue using their current provider.").

### Verified, not rebuilt
- `CreateSubscriptionRequest` carries no provider-settable field — confirmed by reading the request/validator; a subscriber cannot choose a provider.
- `IPaymentMethodSetupRequestFactoryResolver.Resolve` already has a clear "unsupported" failure path (`PaymentMethodSetupService`) for a null factory — registering the new Adyen factory closes that gap without needing new error-handling code.
- `CheckoutApiStoredPaymentChargeProviderGateway` and its resolver already route `ADYEN-ONLINE` — read directly, not re-tested exhaustively (see "not run locally" below).

### Explicitly deferred / not done
- Step 7 of the plan (Payment Providers UI checklist additions — webhook URL display, full Adyen field checklist on the create/update forms) was **not implemented** in this PR; ran out of budget. The merchant-profile page's readiness cards already surface *that* something is missing and link to the Payment Providers page, but the create/update forms themselves were not audited/extended.
- `SubscriptionResponse.ProviderName` is only populated at the checkout/creation response paths, not universally.

## Uncertainty called out in code

`AdyenSetupSessionRequestFactory`'s doc comment flags explicitly: whether Adyen's Sessions API actually accepts `amount.value: 0` for every payment method was **not verified against a live Adyen sandbox**. The zero-amount pattern mirrors Stripe's dedicated `setup` mode and `AdyenInitiationRequestFactory`'s existing request shape as closely as possible, but this is the one place in the PR where real Adyen API behavior is asserted by inference rather than by observation.

## Test plan

- [x] `dotnet build server/Api/Api.csproj --no-incremental` — 0 errors.
- [x] `dotnet test --filter "FullyQualifiedName!~Integration"` from `server/` — 3376 passed, 4 failed, 1 skipped. The 4 failures are the pre-existing `SubscriptionSimulationServiceTests` / `SubscriptionSimulationGuardTests` (x2) / `SubscriptionSimulationDataConsoleServiceTests` failures, unrelated to any file this PR touches (confirmed via `git log origin/dev` on those files).
- [x] `npm run type-check` from `client/` — 0 errors.
- [x] New unit tests: readiness service (all five outcomes, both providers), Adyen setup-session factory (request shape, zero amount, recurring model, shopper reference), merchant-profile service (legacy-resolves-to-Stripe, refusal on not-ready provider, audit event on change), subscription creation (Stripe/Adyen-selected routing, pre-persist refusal, write-free preview).
- [ ] **Not run locally**: a live Adyen sandbox test-merchant lifecycle, and the Mongo integration test suite (both need infrastructure this sandbox doesn't have — no live Adyen credentials, no live mongod). No new integration tests were added in this PR; this is flagged rather than claimed as covered.

### DIFF
--- a/client/app/cross-modules/utilities/subscription-simulation/components/current-subscription-card.test.tsx
+++ b/client/app/cross-modules/utilities/subscription-simulation/components/current-subscription-card.test.tsx
@@ -35,6 +35,7 @@ const baseSubscription: SimulatedSubscription = {
   hasPaymentMethod: null,
   meters: [],
   version: 1,
+  providerName: null,
 };
 
 const noop = () => undefined;
@@ -263,5 +264,20 @@ describe("CurrentSubscriptionCard scheduled plan change", () => {
     renderCard(baseSubscription);
 
     expect(screen.queryByTestId("pending-plan-change")).not.toBeInTheDocument();
+  });
+});
+
+describe("CurrentSubscriptionCard provider name", () => {
+  it("shows the frozen provider a subscription charges through, mapped to a friendly label", () => {
+    renderCard({ ...baseSubscription, providerName: "ADYEN-ONLINE" });
+
+    expect(screen.getByText("Adyen")).toBeInTheDocument();
+  });
+
+  it("shows nothing when the response had no billing account loaded to read it from", () => {
+    renderCard({ ...baseSubscription, providerName: null });
+
+    expect(screen.queryByText("Adyen")).not.toBeInTheDocument();
+    expect(screen.queryByText("Stripe")).not.toBeInTheDocument();
   });
 });
--- a/client/app/cross-modules/utilities/subscription-simulation/components/current-subscription-card.tsx
+++ b/client/app/cross-modules/utilities/subscription-simulation/components/current-subscription-card.tsx
@@ -1,4 +1,5 @@
 import { AlertCircle, CalendarClock, CreditCard, ExternalLink, History, Inbox } from "lucide-react";
+import { Badge } from "@/components/ui-kits/badge/badge";
 import { Button } from "@/components/ui-kits/button/button";
 import { Card } from "@/components/ui-kits/card/card";
 import { Skeleton } from "@/components/ui-kits/skeleton/skeleton";
@@ -11,6 +12,15 @@ import { SubscriptionStatusBadge } from "./subscription-status-badge";
 
 const formatDate = (isoDate: string | null) =>
   isoDate ? new Date(isoDate).toLocaleString() : "—";
+
+// Debugging aid, not customer-facing marketing: an operator comparing two subscriptions that
+// charge through different providers has no other way to see which is which from this screen.
+const PROVIDER_LABELS: Record<string, string> = {
+  STRIPE: "Stripe",
+  "ADYEN-ONLINE": "Adyen",
+};
+const formatProviderName = (providerName: string | null) =>
+  providerName ? (PROVIDER_LABELS[providerName] ?? providerName) : null;
 
 const describeTier = (tier: QuantityDiscountTier) => {
   const range =
@@ -125,6 +135,11 @@ export const CurrentSubscriptionCard = ({
           <div className="flex items-center gap-2">
             <h3 className="font-semibold">{subscription.planName}</h3>
             <SubscriptionStatusBadge status={subscription.status} />
+            {formatProviderName(subscription.providerName) && (
+              <Badge variant="secondary" className="font-normal">
+                {formatProviderName(subscription.providerName)}
+              </Badge>
+            )}
           </div>
           <p className="text-xs text-muted-foreground">
             {scopeLabel} · {subscription.planCode} ·{" "}
--- a/client/app/cross-modules/utilities/subscription-simulation/models/subscription-simulation.model.ts
+++ b/client/app/cross-modules/utilities/subscription-simulation/models/subscription-simulation.model.ts
@@ -115,6 +115,13 @@ export interface SimulatedSubscription {
    */
   meters: MeterTerms[];
   version: number;
+  /**
+   * The payment provider this subscription's billing account was pinned to at creation --
+   * `"STRIPE"` or `"ADYEN-ONLINE"`. Frozen for the subscription's whole life: a later change to
+   * the tenant's merchant-profile selection never moves this. Null wherever the server had no
+   * billing account loaded to read it from.
+   */
+  providerName: string | null;
 }
 
 /** One meter's terms as the subscriber actually bought them. */
--- a/client/app/cross-modules/utilities/subscription/models/subscription-billing.model.ts
+++ b/client/app/cross-modules/utilities/subscription/models/subscription-billing.model.ts
@@ -54,6 +54,27 @@ export interface UpdateBillingProfileRequest {
  * under the deployment-wide fallback — which every other tenant is also using, and which is the one
  * thing a console showing these values has to make visible.
  */
+/** Names the two providers this build can route a subscription charge through. */
+export type SubscriptionPaymentProviderName = "STRIPE" | "ADYEN-ONLINE";
+
+/**
+ * Whether a provider is actually usable for a new subscription right now — see the server's
+ * `SubscriptionPaymentProviderReadiness` enum, which this mirrors exactly.
+ */
+export type SubscriptionPaymentProviderStatus =
+  | "Ready"
+  | "Unsupported"
+  | "NotConfigured"
+  | "Disabled"
+  | "Misconfigured"
+  | "CredentialsUnavailable";
+
+/** One provider's readiness, as rendered on the merchant-profile page's two selection cards. */
+export interface SubscriptionMerchantProfilePaymentProvider {
+  name: SubscriptionPaymentProviderName;
+  status: SubscriptionPaymentProviderStatus;
+}
+
 export interface SubscriptionMerchantProfile {
   legalName: string;
   displayName?: string | null;
@@ -71,6 +92,15 @@ export interface SubscriptionMerchantProfile {
   missingFields: string[];
   isInheritedFromConfiguration: boolean;
   lastUpdatedDateUtc?: string | null;
+  /**
+   * The provider new subscriptions will be routed through — the stored value, or `STRIPE` for a
+   * tenant that has never saved one.
+   */
+  paymentProviderName: SubscriptionPaymentProviderName;
+  /** Whether {@link paymentProviderName} is actually usable for a new subscription right now. */
+  paymentProviderStatus: SubscriptionPaymentProviderStatus;
+  /** Readiness for every provider this build supports, independent of which one is selected. */
+  paymentProviders: SubscriptionMerchantProfilePaymentProvider[];
 }
 
 export interface UpdateMerchantProfileRequest {
@@ -85,6 +115,8 @@ export interface UpdateMerchantProfileRequest {
   primaryColor?: string | null;
   /** A six-digit hex color, with or without the leading '#'. Null clears it back to the default. */
   accentColor?: string | null;
+  /** Required: the console never submits a blank value. */
+  paymentProviderName: SubscriptionPaymentProviderName;
 }
 
 export type FinancialDocumentType = "Invoice" | "TrialInvoice" | "CreditNote";
--- a/client/app/cross-modules/utilities/subscription/pages/subscription-merchant-profile-page.test.tsx
+++ b/client/app/cross-modules/utilities/subscription/pages/subscription-merchant-profile-page.test.tsx
@@ -41,6 +41,12 @@ const profile = (
   missingFields: [],
   isInheritedFromConfiguration: false,
   lastUpdatedDateUtc: "2026-08-25T10:00:00Z",
+  paymentProviderName: "STRIPE",
+  paymentProviderStatus: "Ready",
+  paymentProviders: [
+    { name: "STRIPE", status: "Ready" },
+    { name: "ADYEN-ONLINE", status: "NotConfigured" },
+  ],
   ...overrides,
 });
 
--- a/client/app/cross-modules/utilities/subscription/pages/subscription-merchant-profile-page.tsx
+++ b/client/app/cross-modules/utilities/subscription/pages/subscription-merchant-profile-page.tsx
@@ -1,6 +1,8 @@
 import { useEffect, useRef, useState } from "react";
 import { AlertCircle, Building2, CheckCircle2, Upload, X } from "lucide-react";
+import { useParams } from "react-router";
 import { useProjectStore } from "@seliseblocks/genesis-os";
+import { Badge } from "@/components/ui-kits/badge/badge";
 import { Button } from "@/components/ui-kits/button/button";
 import { Card } from "@/components/ui-kits/card/card";
 import { Input } from "@/components/ui-kits/input/input";
@@ -20,7 +22,41 @@ import {
   useMerchantProfile,
   useUpdateMerchantProfile,
 } from "../hooks/use-merchant-profile";
-import type { SubscriptionMerchantProfile } from "../models/subscription-billing.model";
+import type {
+  SubscriptionMerchantProfile,
+  SubscriptionMerchantProfilePaymentProvider,
+  SubscriptionPaymentProviderName,
+} from "../models/subscription-billing.model";
+
+const PAYMENT_PROVIDERS: {
+  name: SubscriptionPaymentProviderName;
+  label: string;
+}[] = [
+  { name: "STRIPE", label: "Stripe" },
+  { name: "ADYEN-ONLINE", label: "Adyen" },
+];
+
+/** A short, human phrase for each readiness outcome — see the server's own enum remarks. */
+const readinessLabel = (
+  status: SubscriptionMerchantProfilePaymentProvider["status"] | undefined,
+): string => {
+  switch (status) {
+    case "Ready":
+      return "Ready";
+    case "NotConfigured":
+      return "Not configured";
+    case "Disabled":
+      return "Disabled";
+    case "Misconfigured":
+      return "Misconfigured";
+    case "CredentialsUnavailable":
+      return "Credentials unavailable";
+    case "Unsupported":
+      return "Unsupported";
+    default:
+      return "Unknown";
+  }
+};
 
 /** Matches the server's own allow-list — see FinancialDocumentLogoResolver.SniffMimeType. */
 const ALLOWED_LOGO_TYPES = ["image/png", "image/jpeg", "image/jpg", "image/svg+xml"];
@@ -56,6 +92,7 @@ interface MerchantForm {
   logoFileId: string;
   primaryColor: string;
   accentColor: string;
+  paymentProviderName: SubscriptionPaymentProviderName;
 }
 
 const emptyForm: MerchantForm = {
@@ -73,6 +110,7 @@ const emptyForm: MerchantForm = {
   logoFileId: "",
   primaryColor: DEFAULT_PRIMARY_COLOR,
   accentColor: DEFAULT_ACCENT_COLOR,
+  paymentProviderName: "STRIPE",
 };
 
 const toForm = (profile: SubscriptionMerchantProfile): MerchantForm => ({
@@ -93,6 +131,7 @@ const toForm = (profile: SubscriptionMerchantProfile): MerchantForm => ({
   // fields are left untouched, not a placeholder standing in for "unset".
   primaryColor: profile.primaryColor ?? DEFAULT_PRIMARY_COLOR,
   accentColor: profile.accentColor ?? DEFAULT_ACCENT_COLOR,
+  paymentProviderName: profile.paymentProviderName ?? "STRIPE",
 });
 
 /**
@@ -108,6 +147,8 @@ export const SubscriptionMerchantProfilePage = () => {
   // it still carries whichever organization the catalogue was being read as, so a detour through
   // here does not silently reset it.
   const subscriptionLink = useSubscriptionLink();
+  const { itemId } = useParams();
+  const paymentProvidersLink = `/app/${itemId ?? ""}/payment/providers`;
   const { data: profile, isLoading, error } = useMerchantProfile();
   const update = useUpdateMerchantProfile();
   const [form, setForm] = useState<MerchantForm>(emptyForm);
@@ -246,12 +287,21 @@ export const SubscriptionMerchantProfilePage = () => {
         logoFileId: form.logoFileId.trim() || null,
         primaryColor: form.primaryColor || null,
         accentColor: form.accentColor || null,
+        paymentProviderName: form.paymentProviderName,
       },
       { onSuccess: () => setSaved(true) },
     );
   };
 
   const saveError = update.error instanceof Error ? update.error.message : null;
+
+  const selectedProviderStatus = profile?.paymentProviders.find(
+    (provider) => provider.name === form.paymentProviderName,
+  )?.status;
+  const selectedProviderNotReady = Boolean(profile) && selectedProviderStatus !== "Ready";
+  const providerChanged = Boolean(
+    profile && form.paymentProviderName !== (profile.paymentProviderName ?? "STRIPE"),
+  );
 
   return (
     <div className="flex flex-col gap-6">
@@ -432,6 +482,64 @@ export const SubscriptionMerchantProfilePage = () => {
         </div>
       </Card>
 
+      <Card className="flex flex-col gap-5 p-5" data-testid="merchant-payment-provider">
+        <div>
+          <h2 className="text-sm font-medium">Subscription payment provider</h2>
+          <p className="text-xs text-muted-foreground">
+            Which provider a new subscription is routed through at creation. Existing subscriptions
+            keep whichever provider they were created with — changing this never moves them.
+          </p>
+        </div>
+
+        <div className="grid gap-3 sm:grid-cols-2">
+          {PAYMENT_PROVIDERS.map((candidate) => {
+            const status = profile?.paymentProviders.find(
+              (provider) => provider.name === candidate.name,
+            )?.status;
+            const ready = status === "Ready";
+            const selected = form.paymentProviderName === candidate.name;
+
+            return (
+              <button
+                key={candidate.name}
+                type="button"
+                data-testid={`merchant-payment-provider-${candidate.name}`}
+                onClick={() => {
+                  setForm((current) => ({ ...current, paymentProviderName: candidate.name }));
+                  setSaved(false);
+                }}
+                className={`flex flex-col gap-2 rounded-md border p-4 text-left transition-colors ${
+                  selected ? "border-primary ring-1 ring-primary" : "border-input"
+                }`}
+              >
+                <div className="flex items-center justify-between">
+                  <span className="font-medium">{candidate.label}</span>
+                  <Badge variant={ready ? "success" : "secondary"}>
+                    {readinessLabel(status)}
+                  </Badge>
+                </div>
+                {!ready && (
+                  <p className="text-xs text-muted-foreground">
+                    Set this up on the{" "}
+                    <a href={paymentProvidersLink} className="underline">
+                      Payment Providers page
+                    </a>{" "}
+                    before selecting it here.
+                  </p>
+                )}
+              </button>
+            );
+          })}
+        </div>
+
+        {providerChanged && (
+          <p className="text-xs text-muted-foreground" data-testid="merchant-payment-provider-changed">
+            This provider will be used for new subscriptions. Existing subscriptions will continue
+            using their current provider.
+          </p>
+        )}
+      </Card>
+
       <Card className="flex flex-col gap-5 p-5">
         <div>
           <h2 className="text-sm font-medium">Invoice branding</h2>
@@ -557,11 +665,16 @@ export const SubscriptionMerchantProfilePage = () => {
         )}
 
         <div className="flex items-center gap-3">
-          <Button onClick={submit} disabled={isLoading || update.isPending}>
+          <Button
+            onClick={submit}
+            disabled={isLoading || update.isPending || selectedProviderNotReady}
+          >
             {update.isPending ? "Saving…" : "Save merchant profile"}
           </Button>
           <span className="text-xs text-muted-foreground">
-            Accepted from the platform console only — an invoice names a seller in law.
+            {selectedProviderNotReady
+              ? "The selected payment provider isn't ready yet — fix its configuration first."
+              : "Accepted from the platform console only — an invoice names a seller in law."}
           </span>
         </div>
       </Card>
--- a/server/Payment.DomainService/Entities/PaymentDetail.cs
+++ b/server/Payment.DomainService/Entities/PaymentDetail.cs
@@ -36,6 +36,39 @@ public sealed class PaymentDetail
     public string? OrganizationId { get; set; }
 
     /// <summary>
+    /// The item id of the exact <see cref="PaymentProvider"/> row the scope-fallback chain
+    /// resolved and executed this payment against -- set once, at initiation, before the
+    /// provider is ever contacted. Distinct from <see cref="OrganizationId"/>, which is the
+    /// request/operation scope rather than the provider configuration that actually answered it.
+    /// </summary>
+    public string? ResolvedProviderId { get; set; }
+
+    /// <summary>
+    /// The real scope of the resolved <see cref="PaymentProvider"/> row: null for a tenant-level
+    /// configuration, an organization id for one scoped to a single organization. Never coerced
+    /// to the caller's own organization -- a null here means "tenant-wide" and must stay null.
+    /// </summary>
+    public string? ResolvedProviderOrganizationId { get; set; }
+
+    /// <summary>
+    /// When a successful (non-refused) authorization event was confirmed for a
+    /// <see cref="Enums.PaymentFlows.PaymentMethodSetup"/> payment. One of the two independent,
+    /// idempotently-recorded signals a card setup needs before it is Ready -- see
+    /// <see cref="Services.PaymentMethodSetupWebhookStateTransitionService"/>'s remarks on the
+    /// two-signal state machine. Null while pending; never cleared once set, and never set by an
+    /// explicit decline, which is a negative signal handled on its own.
+    /// </summary>
+    public DateTime? SetupAuthorizationConfirmedAtUtc { get; set; }
+
+    /// <summary>
+    /// When a recurring token was confirmed for this setup, whether it arrived inline on the
+    /// authorization event or -- the documented shape -- on a separate, later
+    /// <c>recurring.token.created</c> webhook. The other of the two independent signals; see
+    /// <see cref="SetupAuthorizationConfirmedAtUtc"/>.
+    /// </summary>
+    public DateTime? SetupTokenConfirmedAtUtc { get; set; }
+
+    /// <summary>
     /// Whether this payment came from the console or from an application. See
     /// <see cref="PaymentOrigins"/>. Null on payments taken before this was recorded, which are
     /// not assumed to be either.
--- a/server/Payment.DomainService/Enums/PaymentStatuses.cs
+++ b/server/Payment.DomainService/Enums/PaymentStatuses.cs
@@ -14,6 +14,15 @@ public static class PaymentStatuses
     public const string PartiallyRefunded = "PARTIALLY_REFUNDED";
     public const string Refunded = "REFUNDED";
 
+    /// <summary>
+    /// A card setup that timed out with one of its two completion signals (see
+    /// <c>PaymentMethodSetupWebhookStateTransitionService</c>) never arriving. Terminal for the
+    /// idempotency key that reserved it, the same way
+    /// <see cref="Refused"/>/<see cref="Cancelled"/>/<see cref="MakePaymentFailed"/> are: a fresh
+    /// attempt needs a fresh key rather than resuming a session that will never complete.
+    /// </summary>
+    public const string Expired = "EXPIRED";
+
     public static readonly string[] All =
     [
         Initiating,
@@ -26,6 +35,7 @@ public static class PaymentStatuses
         Captured,
         Cancelled,
         PartiallyRefunded,
-        Refunded
+        Refunded,
+        Expired
     ];
 }
--- a/server/Payment.DomainService/Models/HostedCheckout/HostedCheckoutSessionRequest.cs
+++ b/server/Payment.DomainService/Models/HostedCheckout/HostedCheckoutSessionRequest.cs
@@ -43,4 +43,12 @@ public sealed class HostedCheckoutSessionRequest
     public string? ShopperEmail { get; set; }
     [JsonPropertyName("shopperInteraction")]
     public string ShopperInteraction { get; set; } = "Ecommerce";
+
+    /// <summary>
+    /// Restricts which payment method types Adyen may present the shopper. Omitted (null) for an
+    /// ordinary priced checkout, which offers everything the merchant account itself allows.
+    /// </summary>
+    [JsonPropertyName("allowedPaymentMethods")]
+    [JsonIgnore(Condition = JsonIgnoreCondition.WhenWritingNull)]
+    public List<string>? AllowedPaymentMethods { get; set; }
 }
--- a/server/Payment.DomainService/Models/Webhooks/ParsedWebhookEvent.cs
+++ b/server/Payment.DomainService/Models/Webhooks/ParsedWebhookEvent.cs
@@ -20,7 +20,20 @@ public sealed class ParsedWebhookEvent
     /// <summary>Provider's own event name, retained for logging and inbox records.</summary>
     public required string EventCode { get; init; }
 
-    public required WebhookIntent Intent { get; init; }
+    /// <summary>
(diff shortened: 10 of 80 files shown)


fahad0208:
Pushed a follow-up commit addressing all 4 P1 and 3 P2 findings plus the CodeQL check: **bb355eab** (on top of e602e4ba).

## P1 fixes

**1. Checkout could validate one merchant account and charge another.**
`ISubscriptionPaymentProviderReadinessService.CheckAsync` now returns `SubscriptionPaymentProviderReadinessResult` (readiness + the actual resolved `PaymentProvider`), not a bare enum. The resolved provider's `OrganizationId` is frozen onto `BillingAccount.ProviderOrganizationId` at creation time (in `SubscriptionCreationService`) and threaded into `MakePaymentRequest.OrganizationId` / `CreatePaymentMethodSetupRequest.OrganizationId` in `SubscriptionCheckoutService`. After the charge/setup returns, its `PaymentResponse.OrganizationId` (new field, mapped from `PaymentDetail.OrganizationId`) is explicitly compared against the frozen scope — a mismatch fails closed with `subscription_payment_provider_scope_mismatch` rather than being adopted silently.

While wiring this I found `PaymentMethodSetupService` ignored `CreatePaymentMethodSetupRequest.OrganizationId` entirely (`CreateRecord` always stamped the caller's ambient `context.OrganizationId`) — the field existed but nothing read it. It now resolves through `IPaymentOrganizationResolver`, the same authorization path `PaymentReservationService` (the charge path) already used, so a setup session opens under the exact scope requested and permitted.

**2. Missing billing account silently fell back to Stripe.** Removed the fallback. `SubscriptionCheckoutService` now returns `subscription_billing_account_provider_unavailable` (`PaymentFailureKind.Unavailable`) and logs tenant/subscription/billing-account/correlation hashes. Test: `An_adyen_subscription_with_a_missing_billing_account_fails_closed_rather_than_falling_back_to_stripe`.

**3. Adyen tokenization used CardOnFile instead of Subscription.** `MakePaymentRequest.RecurringModel` (the field already existed, but `MakePaymentRequestValidator` unconditionally rejected it — `.Empty()` — since it was meant to be pipeline-managed) is now allowed to be `"Subscription"` specifically; subscription checkout's initial charge sets it. `AdyenInitiationRequestFactory.Create` honors it when present, defaulting to `"CardOnFile"` for any other caller (checked — no other caller currently sets `SavePaymentMethod`+recurring intent, so this is additive). `AdyenSetupSessionRequestFactory` now hardcodes `"Subscription"` since the subscription module is its only caller.

**Not verified live**: I don't have Adyen sandbox credentials in this environment. This follows https://docs.adyen.com/online-payments/tokenization/make-token-payments (Subscription = merchant-initiated fixed-schedule; CardOnFile = shopper-initiated), but the actual API acceptance/behavior was not exercised against a live merchant account. Flagged in code comments at both factories.

**4. Zero-value setup didn't restrict payment methods; consent-decline gap.** `HostedCheckoutSessionRequest` gained `AllowedPaymentMethods`; `AdyenSetupSessionRequestFactory` sets it to `["scheme"]` (cards — Adyen's documented generic card type, not live-verified). Also found and fixed a related bug while checking the webhook path: `PaymentMethodSetupWebhookStateTransitionService` marked a setup `Authorized` whenever the provider reported success, even when `ApplyAuthorisationTokenAsync` found no token/shopper-reference (i.e., the shopper declined the `askForConsent` prompt) and stored nothing — leaving a payment that looked terminally successful with no card ever saved. It now downgrades that specific shape (success=true, no token, no shopper reference) to a refusal, reaching `Refused` instead.

## P2 fixes

**5. Preview didn't enforce readiness.** `SubscriptionCreationService` now runs the readiness check on both `preview` and real creation, returning the same named error from both. Preview still writes nothing. Parity test: `Preview_is_refused_with_the_same_named_error_as_creation_when_the_provider_is_not_ready` (all 4 non-Ready states).

**6. `/current` and other responses didn't reliably carry `ProviderName`.** Added `SubscriptionResponseMapperExtensions.ToResponseAsync` — fetches the billing account once and derives `ProviderName` + `HasPaymentMethod` together — and made it the path every `SubscriptionResponse`-producing call site in `SubscriptionCheckoutService`, `SubscriptionCancellationService`, and `SubscriptionPlanChangeService` goes through (current/pending/unpaid/resume/cancel/plan-change), replacing the previously-optional, independently-suppliable `providerName`/`hasPaymentMethod` parameters that several call sites had simply omitted. `QuantityChangeResponse` also gained `ProviderName`, threaded through a new `DescribeAsync` (all 6 call sites updated).

**7. Readiness reimplemented a simplified first-match resolution instead of reusing real charge-time resolution.** `SubscriptionPaymentProviderReadinessService` now resolves via `IPaymentRepository.GetProviderAsync` — the exact enabled-filtered, scope-chain-fallback lookup a real charge uses — rather than its own `GetProvidersAsync`+first-match logic. `GetProvidersAsync` is still used, but only afterward and only to distinguish `Disabled` from `NotConfigured` for the error message; it never participates in the Ready decision. Regression test (`An_organization_specific_disabled_configuration_still_resolves_ready_through_a_tenant_wide_fallback`) uses a hand-rolled `IPaymentRepository` that reimplements the real scope-fallback semantics in memory, so it would fail if readiness ever reverted to picking only the first scope match.

## CodeQL

`gh pr checks 393` showed `CodeQL fail` despite all three `Analyze` jobs passing — **not flaky**: the aggregate check carried one real finding (`clear text storage of sensitive information`, high severity) in `PaymentStateTransitionService.cs:156`, a structured-log call passing the raw `payment.ProviderName` instead of going through this module's own `PaymentLogValue.Label(...)` sanitizing wrapper (everywhere else in the payment module does). Fixed by routing it through `PaymentLogValue.Label`. Provider name itself isn't a secret (`STRIPE`/`ADYEN-ONLINE`), but the fix brings the line in line with the module's established logging convention and should satisfy the query. Will confirm on the next CodeQL run against this commit.

## Deferred (unchanged from the review, explicitly not attempted this round)

- Adyen readiness checklist / webhook setup guidance in the Payment Providers UI
- Mongo integration tests for merchant-profile persistence and provider-scoped billing accounts
- Real Adyen sandbox lifecycle verification (no credentials available in this environment)
- Published subscription integration documentation
- Full provider-routing test matrix across renewal/dunning/plan-change/quantity-change/overage/refunds/invoices

## Verification

- `dotnet build server/Api/Api.csproj --no-incremental` — **0 errors** (3906 pre-existing warnings, none new from this change).
- `dotnet test --filter "FullyQualifiedName!~Integration"` (from `server/`) — **3390 passed**, 1 skipped (pre-existing, unrelated PDF engine test), **4 failed**, all confirmed pre-existing and unrelated by running the exact same tests against `origin/dev`: `SubscriptionSimulationServiceTests.Refuses_the_console_without_the_simulation_permission`, `SubscriptionSimulationGuardTests.Refuses_the_console_without_the_permission`, `SubscriptionSimulationGuardTests.Refuses_the_console_with_no_permissions_at_all`, `SubscriptionSimulationDataConsoleServiceTests.Find_refuses_the_console_without_the_simulation_permission` (all pre-existing NullReferenceException/authorization bugs unrelated to this PR's diff — none of the touched files appear in `git diff origin/dev` for the Simulation module). One additional test (`FinancialDocumentRendererHealthMonitorTests.Keeps_reprobing_on_the_configured_interval_while_unhealthy`, a timing-sensitive Worker test, also untouched by this diff) failed once and passed cleanly on an isolated rerun — flaky, not a regression.
- `npm run type-check` (from `client/`) — **0 errors**. No client code references `providerName` yet, so this round's server-side changes had nothing to break there.

Added/updated tests across `SubscriptionCheckoutServiceTests`, `SubscriptionCreationServiceTests`, `SubscriptionPaymentProviderReadinessServiceTests`, `AdyenInitiationRequestFactoryTests`, `AdyenSetupSessionRequestFactoryTests`, `ZeroAmountCardSetupTests`, plus mock/constructor updates in cancellation-related test files for the new `IBillingAccountRepository` dependency and the readiness result type change.

fahad0208:
Fixed both new P1s in commit 4fa65184. Details, deferrals, and verification results below.

## Finding 1 — provider-scope comparison was comparing the wrong concept

Root cause confirmed as described: `PaymentResponse.OrganizationId` is the *request/operation scope* (mapped straight from `PaymentDetail.OrganizationId`), not the identity of the `PaymentProvider` row the scope-fallback chain actually resolved. Comparing it against the billing account's frozen `ProviderOrganizationId` produced false-positive `subscription_payment_provider_scope_mismatch` failures whenever resolution correctly fell through to a shared tenant/console-level configuration, and it only fired *after* the Adyen session and payment record already existed.

Changes:
- Added `BillingAccount.ProviderId` — the exact `PaymentProvider.ItemId` readiness resolved, frozen alongside the existing `ProviderOrganizationId` in `SubscriptionCreationService`.
- Added `PaymentDetail.ResolvedProviderId` / `ResolvedProviderOrganizationId`, persisted by `PaymentRepository.SaveInitiationRequestAsync` at the exact point `HostedCheckoutInitiationService` / `PaymentMethodSetupService` resolve the `PaymentProvider` row — before the provider is contacted. `ResolvedProviderOrganizationId` preserves `null` for a tenant-level row rather than coercing it, unlike the old `?? context.OrganizationId` pattern.
- Added `MakePaymentRequest.ExpectedProviderId` (`[JsonIgnore]`, so it can never be set from the public HTTP body) and `CreatePaymentMethodSetupRequest.ExpectedProviderId`. `SubscriptionCheckoutService` passes the billing account's frozen `ProviderId` through both. `HostedCheckoutInitiationService`/`PaymentMethodSetupService` now refuse to proceed — **before building the provider request, before calling the session client** — if the resolved row doesn't match the expectation, via a new `payment_provider_scope_mismatch` failure. This is the fail-closed, before-the-external-call fix the finding asked for.
- `HostedCheckoutInitiationService.RecoverAsync` got the same guard for a resumed initiation.
- `SubscriptionCheckoutService`'s own post-hoc check is kept as defense in depth (for accounts with no frozen `ProviderId`, i.e. created before this change, where the payment module's check has nothing to compare against and is skipped) — now comparing `ResolvedProviderId` against `BillingAccount.ProviderId`, never organization ids.
- Corrected `PaymentResponse.OrganizationId`'s doc comment, which previously and incorrectly described it as the resolved provider's scope.

Deliberately out of scope: the equivalent check was **not** added to renewal, plan-change, quantity-change, or usage-invoice charging (`SubscriptionRenewalService`, `SubscriptionPlanChangeService`, `SubscriptionQuantityChangeService`, `SubscriptionUsageRatingProcessor`) — none of these were named in the finding, and extending it there would have widened this round considerably. They carry the same theoretical exposure and are worth a follow-up look.

## Finding 2 — Adyen setup webhook never reached the setup handler

Confirmed: Adyen's `AUTHORISATION` event is identical for a real payment and a zero-value setup — no event-name-level distinction, unlike Stripe's `setup_intent.succeeded`. `AdyenWebhookNormalizer` has no access to the loaded `PaymentDetail` at parse time (it's pure parsing, no DB access), and `PaymentWebhookStateTransitionService` only sees `webhook.Intent` at dispatch time, before it loads the payment for the `Authorization` case. `PaymentWebhookIntakeService.VerifyPaymentOwnershipAsync`, however, already loads and verifies ownership of the `PaymentDetail` in order to route at all — so that's where the fix landed.

- `ParsedWebhookEvent.Intent` changed from `init` to a settable property (documented why).
- After ownership verification, if `webhookEvent.Intent == Authorization` and the loaded payment's `PaymentFlow == PaymentFlows.PaymentMethodSetup`, the intent is corrected to `PaymentMethodSetup` before the event is stored in the inbox. This is provider-agnostic (keyed off `PaymentFlow`, not provider name) but is a no-op for Stripe, whose normalizer already emits `PaymentMethodSetup` intent directly for its own setup events — verified by reading `StripeWebhookNormalizer`.
- `PaymentWebhookStateTransitionService`'s existing `case WebhookIntent.PaymentMethodSetup` dispatch needed no changes — it now receives the corrected intent for Adyen too.

## Finding 3 — missing-token-on-authorization ≠ declined consent

Per Adyen's tokenization docs, the recurring token is asynchronous and delivered on a separate `recurring.token.created` webhook that can arrive before or after the `AUTHORISATION` webhook. The prior round's fix ("no token in this event ⇒ treat as Refused") conflated "declined" with "token hasn't arrived on this event yet" — a real false-positive.

Replaced with the two-signal state machine from the finding's spec:
- `PaymentDetail.SetupAuthorizationConfirmedAtUtc` / `SetupTokenConfirmedAtUtc` — two independent, idempotently-recorded facts (`IPaymentRepository.TryRecordSetupAuthorizationConfirmedAsync` / `TryRecordSetupTokenConfirmedAsync`, first-write-wins via a null-filtered Mongo update).
- `PaymentMethodSetupWebhookStateTransitionService.ApplyAsync`: a successful `AUTHORISATION` records the authorization signal only (an inline token, if Adyen ever sends one this way — not verified against a live sandbox either way — is treated as an opportunistic fast path, never a requirement) and then checks readiness.
- `StoredPaymentMethodLifecycleService.ApplyTokenEventAsync`: for a setup-flow correlated payment, no longer requires `PaymentStatus == Authorized` before proceeding (that status is now withheld until both signals land, so requiring it would deadlock a token-first delivery); stores the card and records the token signal, then checks readiness.
- `PaymentMethodSetupCompletion` (new, internal, static — deliberately not a third injected service, to avoid a DI cycle between the two callers, which already inject each other's collaborators): completes the setup — flips status, publishes `PaymentMethodSetupSucceeded` — only once both signals are present, from whichever call reaches it second. Idempotent via `ApplyAuthorisationAsync`'s existing deduplication-key check (a fixed, event-independent key: `{paymentId}:{eventType}:setup-ready`), so a duplicate delivery or two calls racing on the second signal cannot double-complete.
- An explicit provider decline/failure/cancellation is still the only thing that produces `Failed`, unconditionally, regardless of the token signal, and a decline recorded first blocks a later-arriving token from resurrecting the setup.

**Deferred, explicitly** (per the finding's own instruction to say so rather than build partially): an **Expired** state for a setup stuck pending one signal past a timeout. I looked for a general "expire a stuck-pending payment" sweep to reuse and didn't find one in this module — the closest is `ISubscriptionWorkScheduler.ScheduleActivationRecoveryAsync`, which watches an *Incomplete subscription*, not a specific payment's missing webhook signal. Building a new timeout mechanism was judged out of scope for this round. A setup stuck on one signal today stays pending indefinitely rather than expiring — documented as such in `PaymentMethodSetupWebhookStateTransitionService`'s own remarks.

**Not independently verified against a live Adyen sandbox** (flagging per the task's own ground rules, not glossing over it): whether Adyen ever reports the recurring token inline on the `AUTHORISATION` event itself (as opposed to only ever on the separate token webhook) is not confirmed either way in this environment — the code treats it as an opportunistic fast path so it's correct either way, but I could not exercise it live.

## Tests added
- `PaymentMethodSetupTests.cs`: successful authorization with no inline token now waits rather than declining; a token signal recorded first lets a later authorization complete the setup; a duplicate authorization delivery does not double-complete (existing decline/expiry tests unmodified and still pass).
- `StoredPaymentMethodLifecycleServiceTests.cs`: token arriving before authorization for a setup-flow payment is stored without throwing and without completing; token arriving after authorization completes the setup; a token arriving for an already-declined setup is ignored rather than resurrecting it.
- `SubscriptionCreationServiceTests.cs` / `SubscriptionCheckoutServiceTests.cs`: `BillingAccount.ProviderId` freezing; a charge resolved under a different provider row than the account was pinned to fails closed; an account with no frozen `ProviderId` (pre-existing accounts) is accepted rather than failing closed against a fact it was never given.

All using mocked/synthetic payloads following this codebase's existing conventions — no live Adyen calls were made or attempted.

## Verification
- `dotnet build server/Api/Api.csproj --no-incremental`: **0 errors**.
- `dotnet test --filter "FullyQualifiedName!~Integration"` (server/): **3399 passed, 0 unexpected failures, 4 failed, 1 skipped** (3404 total). The 4 failures are all `SubscriptionSimulationGuardTests` / `SubscriptionSimulationServiceTests` / `SubscriptionSimulationDataConsoleServiceTests` — confirmed pre-existing and unrelated: `git diff origin/dev` shows zero changes to any of those test files on this branch.
- `npm run type-check` (client/): **0 errors** (no-op as expected — this round is server-only).
- CI on this push (commit 4fa65184): **all green** — Analyze (csharp/javascript-typescript/actions), CodeQL, mongo-integration, validate-client, repo-hygiene, and initialization all `pass`. No regressions.

## Remaining deliberate deferrals (carried from the last round, plus this round's addition)
- Payment Providers UI readiness checklist.
- Purpose-built Mongo integration coverage for profile/billing-account provider persistence.
- Published integration-guide docs.
- Cross-operation provider-routing test matrix (i.e. extending finding 1's fix to renewal/plan-change/quantity-change/usage-invoice charging — see above).
- Real Adyen sandbox lifecycle verification (setup two-signal ordering, whether the token ever arrives inline on `AUTHORISATION`).
- New this round: an `Expired` timeout state for a setup stuck pending one signal (finding 3) — no existing sweep to reuse; would need its own mechanism, judged out of scope.

Commit: 4fa65184

fahad0208:
Pushed a follow-up commit addressing this round's two P1s and the P2: `e651b753` (on top of `4fa65184`).

## Finding 1 (P1) - token processing crash-replay gap - fixed

`StoredPaymentMethodLifecycleService.ApplyTokenEventAsync` now resolves the correlated payment by `payload.EventId` unconditionally, before the `existing == null` branch, instead of only inside it. Concretely:

- The `existing == null` branch keeps its original validation and upsert logic completely unchanged (still gates on `RememberCard`, shopper match, and the setup-flow `Authorized`-deferral rule; still returns early on an already-declined setup).
- A new `else if` arm now runs when the stored method already exists and the correlated payment is a setup flow whose shopper matches: it still calls `RecordSetupTokenSignalAsync` (which calls `TryRecordSetupTokenConfirmedAsync` then `PaymentMethodSetupCompletion.TryCompleteAsync`), even though there's nothing new to upsert. An already-declined setup is still logged and skipped, matching the `existing == null` path's behavior.

This closes both scenarios from the finding: (a) a crash between the upsert and `TryRecordSetupTokenConfirmedAsync`, replayed by Adyen's at-least-once delivery, now still records the signal on retry; (b) a shopper who already has a stored method from an unrelated prior setup no longer short-circuits correlation for a brand-new setup.

Added two tests to `StoredPaymentMethodLifecycleServiceTests`: `Token_created_retry_after_a_crash_still_records_the_setup_signal` (stored method pre-exists, `SetupTokenConfirmedAtUtc` still unset -- asserts the retry writes it) and `Token_reused_from_a_prior_stored_method_still_correlates_the_new_setup` (shopper's existing stored method predates this setup entirely -- asserts the new setup's signal is still recorded, not the old one's).

## Finding 2 (P1) - legacy billing accounts never get a provider identity - fixed

`BillingAccountRepository.GetOrCreateAndReconcileAsync` now calls a new private `BackfillProviderIdentityAsync` after its existing `FindOneAndUpdateAsync`/duplicate-key-retry logic. It backfills `ProviderId`/`ProviderOrganizationId` only when the stored document's `ProviderId` is currently null and the caller supplied one, via a conditional `UpdateOneAsync` filtered on `ProviderId == null` -- the same first-write-wins compare-and-set convention already used by `PaymentRepository.TryRecordSetupTokenConfirmedAsync`/`TryRecordSetupAuthorizationConfirmedAsync`. A non-null `ProviderId` is never touched, so a frozen identity can never be silently moved by an ordinary reconcile call. This was kept out of the existing `Reconciliation()` find-and-modify because `$set`/`$setOnInsert` can't express "only if currently null" -- it needs its own conditional write.

Added two integration tests to `SubscriptionRepositoryIntegrationTests` (real Mongo, so the CAS behavior is actually exercised rather than mocked): `A_legacy_account_with_no_provider_identity_is_backfilled_on_reconcile` and `An_already_frozen_provider_identity_is_never_overwritten_by_reconcile`. Both passed in CI's `mongo-integration` job.

## Finding 3 (P2) - no terminal recovery for a missing webhook signal - implemented, not deferred

Checked first: `PaymentReconciliationBackgroundService`/`PaymentReconciliationWorkHandler` already run a per-tenant recovery-processor pattern (`IPaymentRecoveryProcessor.RecoverStaleAsync`, `IStoredPaymentMethodRemovalRecoveryProcessor.RecoverDueRemovalsAsync`, etc.) in both the legacy direct-sweep and durable-queue paths, so I hooked into that rather than building new scheduling infrastructure.

- Added `PaymentStatuses.Expired`, a new `PaymentMethodSetupTimeoutSeconds` option (default 1800s), and two `IPaymentRepository` methods: `GetDueSetupExpiryCandidatesAsync` (setup flow, still `Processing`, `CreatedAtUtc` past the cutoff, either completion signal null) and `TryExpireSetupAsync` (compare-and-set on `PaymentStatus == Processing`, so a concurrent completion or decline always wins over the sweep).
- New `PaymentMethodSetupExpiryProcessor` (`IPaymentMethodSetupExpiryProcessor.ExpireDueAsync`) runs the sweep, logs `AgeSeconds`/`MissingSignal` (`authorization`, `token`, or `both`) per expiry, and records two new `PaymentWorkMetrics` instruments: `payment.setup.expired` (counter, tagged `missing_signal`) and `payment.setup.pending_age` (histogram, seconds, same tag) -- recorded for every candidate the sweep looks at, not only the ones it expires, so age is visible climbing before the timeout hits.
- Wired into both `PaymentReconciliationWorkHandler.ExecuteAsync` (durable queue) and `PaymentReconciliationBackgroundService.ReconcileTenantAsync` (legacy direct sweep).
- `PaymentMethodSetupService`'s idempotency-key replay check now treats `Expired` like the existing `Refused`/`Cancelled`/`MakePaymentFailed` terminal group -- returns `PaymentFailureKind.Timeout` / `payment_method_setup_expired` rather than resuming a session that will never complete, so a fresh attempt needs a fresh idempotency key, consistent with how the other terminal statuses already behave.

Added `PaymentMethodSetupExpiryProcessorTests` (5 cases: no candidates, expire-past-timeout with both signals missing, lost-race-does-not-count-as-expired, expire with only the token signal missing, and the repository is queried with the configured cutoff).

One caveat worth flagging explicitly: I did not add outbox/downstream notification for an expired setup (e.g., no `PaymentMethodSetupExpired` event consumed by subscription activation). The finding only asked for a terminal status plus recovery plus observability, and adding a new cross-domain event/consumer felt like meaningfully larger scope than this round's two P1s -- happy to add it as a follow-up if the reviewer wants subscription-side awareness of an expired setup rather than just a fresh checkout attempt working cleanly afterward.

## Build / test / type-check / CI

- `dotnet build server/Api/Api.csproj --no-incremental` -- 0 errors (warnings only, pre-existing).
- `dotnet test --filter "FullyQualifiedName!~Integration"` (from `server/`) -- 3406 passed, 4 failed, 1 skipped, 3411 total. The 4 failures are all in `SubscriptionSimulationGuardTests`/`SubscriptionSimulationServiceTests`/`SubscriptionSimulationDataConsoleServiceTests` -- confirmed pre-existing via `git diff origin/dev` showing zero diff on those test files or the `Subscription.DomainService/Simulation/` production code they exercise.
- `npm run type-check` (from `client/`) -- 0 errors (no-op, as expected for a server-only round).
- CI on commit `e651b753`: all checks green, including `CodeQL`, both `mongo-integration` shards, `validate-client`, `Analyze (csharp/javascript-typescript/actions)`, `repo-hygiene`.

## Still deliberately deferred (unchanged from prior rounds, plus nothing new this round)

- Payment Providers UI checklist.
- Purpose-built Mongo integration coverage beyond what's already in `SubscriptionRepositoryIntegrationTests` (the two new Finding 2 tests use that existing harness; no new integration test infrastructure was built).
- Integration-guide docs.
- Cross-operation provider-routing test matrix.
- Real Adyen sandbox verification (not possible in this environment -- everything above is unit/Mongo-integration coverage against synthetic/mocked webhook payloads following this codebase's existing conventions).
- The setup-expiry-to-downstream-notification gap noted above under Finding 3 (net new scope, not a repeat of a prior deferral).

Flagging per the task's request for explicit uncertainty: I did not find or infer any Adyen-specific webhook semantics beyond what the two-signal model already established in the prior round (`SetupAuthorizationConfirmedAtUtc`/`SetupTokenConfirmedAtUtc`); Finding 1's fix is purely about when the existing correlation/signal-recording code runs, not about any new webhook field or semantic, so no new "genuinely uncertain" comment was needed in the code for it.

fahad0208:
Pushed 4223c540, addressing all three findings from this review round.

## Finding 1 (P1) — expiry/completion race, fixed

`PaymentRepository.TryExpireSetupAsync`'s compare-and-set filtered only on `Status == Processing`. That's not sufficient: the token or authorization webhook can record its signal (`TryRecordSetupTokenConfirmedAsync` / `TryRecordSetupAuthorizationConfirmedAsync`) *after* the sweep read its candidate list but *before* its expiry write runs, while the completion that signal unlocks is still in flight (or waiting on a retry). Status stays `Processing` through that whole window, so the old filter let the sweep win a race it should always lose.

Fix: the filter now re-checks, atomically in the same update, that a signal is still missing (`SetupAuthorizationConfirmedAtUtc == null || SetupTokenConfirmedAtUtc == null`) — the identical condition `GetDueSetupExpiryCandidatesAsync` uses at selection time, now also enforced at the point of the write itself.

Verified the completion-retry claim rather than assuming it: both `PaymentMethodSetupWebhookStateTransitionService.ApplyAsync` and `StoredPaymentMethodLifecycleService.RecordSetupTokenSignalAsync` already call `PaymentMethodSetupCompletion.TryCompleteAsync` unconditionally after recording their signal, on every delivery including a redelivery — so the ordinary "signal arrived, but the same call's completion hasn't been retried yet" case self-heals on the next webhook delivery. The real residual gap is a process crashing between recording the final signal and calling `TryCompleteAsync`, with no further webhook redelivery to retry it. Added `IPaymentRepository.GetPendingSetupsAsync` (all currently-`Processing` setups for a tenant, any age, any signal state) and wired the expiry sweep (`PaymentMethodSetupExpiryProcessor`) to call `TryCompleteAsync` on any pending setup that already has both signals — so that case is now also self-healed by the sweep instead of staying stuck once the CAS fix correctly refuses to expire it.

New Mongo integration tests in `PaymentRepositoryIntegrationTests`: `TryExpireSetup_loses_the_race_to_a_signal_recorded_after_the_candidate_was_read` reproduces the exact race (read candidate → record the missing signal → attempt expiry → assert it's refused and status stays `Processing`), plus a companion test confirming a genuinely-missing-signal setup still expires normally, and a test for `GetPendingSetupsAsync`'s filtering.

## Finding 2 (P2) — losing backfill writer, fixed

`BillingAccountRepository.BackfillProviderIdentityAsync` used `UpdateOneAsync` and, on `ModifiedCount == 0` (lost the race to a concurrent backfill), returned the stale pre-update `stored` object — still `ProviderId == null` — rather than what actually landed. Switched to `FindOneAndUpdateAsync` with `ReturnDocument.After`: the call that wins the race gets the document its own write produced, and a call whose filter no longer matches (lost the race) falls back to a fresh `FindAsync` read rather than the stale in-memory value, so every caller ends up returning the same non-null winning `ProviderId`.

New concurrent Mongo integration test in `SubscriptionRepositoryIntegrationTests`: `Concurrent_backfills_of_the_same_legacy_account_agree_on_one_non_null_provider_id` fires two `GetOrCreateAndReconcileAsync` calls at the same legacy (null-`ProviderId`) account via `Task.WhenAll` and asserts both results carry the same non-null `ProviderId`, matching what's actually persisted.

## Finding 3 (P3) — telemetry and doc comment, fixed (option a)

Went with the broader fix rather than just the description: `payment.setup.pending_age` previously only ever observed setups already selected as expiry candidates (i.e. already past the timeout), so it could never show age climbing over time. It's now recorded for every setup `GetPendingSetupsAsync` returns, at whatever age, tagged by missing signal. Updated the histogram's own `description` to match. Confirmed no gauge/counter API mismatch — it was already a `Histogram<double>` with `.Record()`, correctly typed; the only issue was what got fed into it.

Also removed the stale doc comment on `PaymentMethodSetupWebhookStateTransitionService` claiming "Expired" was unimplemented and setups stay pending indefinitely — no longer true since the prior round's expiry sweep. Replaced with an accurate `<item>` describing `Expired` and pointing at `PaymentMethodSetupExpiryProcessor` and this round's CAS fix.

## Results

- Build: `dotnet build server/Api/Api.csproj --no-incremental` — 0 errors.
- Tests: `dotnet test --filter "FullyQualifiedName!~Integration"` from `server/` — 3407-3408/3413 passed across two runs (one flaky `PeriodicPingBackgroundServiceTests` case that passes standalone and in the second full run), remaining failures are the four pre-existing `SubscriptionSimulation*`/`SubscriptionSimulationGuardTests` failures — confirmed via `git diff origin/dev` that none of those files differ from `origin/dev`, so they predate this PR entirely.
- Client type-check: `npm run type-check` — 0 errors (server-only round, no-op as expected).
- CI on 4223c540: all required checks green, including both `mongo-integration` jobs (which is where the new Finding 1/2 integration tests actually ran against real MongoDB) and CodeQL.

## Still deliberately deferred from prior rounds (unchanged)

- Payment Providers UI checklist (frontend surface for provider selection/readiness) not built this round.
- Further purpose-built Mongo integration coverage beyond what's been added across rounds — this round added targeted races for Findings 1 and 2 specifically, not a general sweep of the module.
- Integration-guide docs for merchants/operators.
- A cross-operation provider-routing test matrix (every payment operation × both providers).
- Real Adyen sandbox verification — the inline-token-on-AUTHORISATION fast path is still unconfirmed against a live sandbox, exactly as noted previously.
- The previously-noted missing outbox event for an expired setup (an `Expired` setup still does not publish anything to the outbox) — not addressed this round; flagging again since it wasn't in scope of these three findings.

fahad0208:
## Fixed: capped query mixing recovery with telemetry (P2, reviewer round 5)

Commit: 1916e699

### What was wrong

`GetPendingSetupsAsync` (added last round) returned only the oldest 1-200 `Processing` setups and fed two unrelated consumers off that one capped result: (1) recovering setups that already had both completion signals but never got their terminal transition applied, and (2) the `payment.setup.pending_age` telemetry. If a tenant had more than one batch's worth of older, genuinely-incomplete setups, a newer setup that was already fully ready to complete could fall outside the oldest-first window and sit stuck until the older backlog drained or expired. The telemetry had the same problem in reverse: it never actually observed "every currently pending setup" as documented, only whatever fit in the first capped batch.

### What changed

- **`IPaymentRepository.GetSetupsReadyForCompletionAsync`** (`server/Payment.DomainService/Repositories/IPaymentRepository.cs`, `PaymentRepository.cs`) — a dedicated query filtered to `PaymentFlow=PaymentMethodSetup`, `PaymentStatus=Processing`, and both `SetupAuthorizationConfirmedAtUtc`/`SetupTokenConfirmedAtUtc` non-null. Not capped by an oldest-first window the way expiry candidates are.
- **`IPaymentRepository.GetPendingSetupAgeSummaryAsync`** — a MongoDB aggregation (`Match` + `Group`, same pattern already used by `PaymentWorkQueue.DescribeDepthAsync`) that computes count and oldest `CreatedAtUtc` per missing-signal category ("authorization"/"token"/"both") across **every** pending setup for the tenant, not a capped batch read into app code. New `PendingSetupAgeSummary` record carries the result.
- **`PaymentMethodSetupExpiryProcessor.ExpireDueAsync`** now runs two independent passes instead of one shared loop: `CompleteReadySetupsAsync` pages through the new ready-query (stops when a page comes back short of the batch size, or a full page completes nothing — a defensive stop, not expected to trigger given `PaymentMethodSetupCompletion`'s own idempotency), and `RecordPendingAgeAsync` publishes the aggregation's oldest-age-per-category directly.
- **`payment.setup.pending_age`** metric description updated (`PaymentWorkMetrics.cs`) to describe what it now actually is: the oldest offender per category from an uncapped Mongo aggregation, not a per-document sample from a capped batch.

### Tests added

- `PaymentRepositoryIntegrationTests.GetSetupsReadyForCompletionAsync_finds_a_ready_setup_behind_a_backlog_larger_than_the_batch_cap` — creates `limit + 3` older setups still missing a signal plus one newer setup with both signals recorded, calls the new query with `limit`, and asserts the ready setup is returned — reproducing the exact starvation scenario the old shared query got wrong.
- `PaymentRepositoryIntegrationTests.GetSetupsReadyForCompletionAsync_returns_only_setups_with_both_signals` and `GetPendingSetupAgeSummaryAsync_aggregates_the_oldest_age_per_missing_signal` — direct coverage of both new queries.
- `PaymentMethodSetupExpiryProcessorTests.Every_ready_setup_is_completed_even_when_more_than_one_batch_worth_exists` — unit test proving the processor pages through multiple ready-to-complete batches within one sweep rather than stopping at the first.
- Updated the two existing processor tests that referenced the old shared method.

### Results

- Build (`dotnet build server/Api/Api.csproj --no-incremental`): **0 errors**
- Test (`dotnet test --filter "FullyQualifiedName!~Integration"` from `server/`): **3409 passed, 4 failed, 1 skipped** out of 3414. The 4 failures are all in `SubscriptionSimulationGuardTests` — the known pre-existing, unrelated failures called out for this task; confirmed no diff to that file or its production code on this branch vs `origin/dev`.
- Client type-check (`npm run type-check` from `client/`): **0 errors** (server-only change, no-op as expected).
- CI on this push: **all green** — `Analyze (actions/csharp/javascript-typescript)`, `CodeQL`, both `mongo-integration` shards, `validate-client`, `initialization`, `repo-hygiene` all passed. (`build-client`/`build-worker`/`sonarqube`/`sca-scan-dotnet`/deploy-preview jobs skipped, as expected for this branch.)

### Deferred items carried forward from prior rounds (unchanged, reviewer already confirmed none are correctness blockers)

- Payment Providers UI checklist (merchant-facing provider selection screen) — not built yet.
- Integration-guide documentation for merchant provider setup — not written yet.
- Cross-operation provider-routing test matrix (every payment operation type × both providers) — only the setup/webhook paths are covered today.
- Real Adyen sandbox verification — all testing to date is against mocks/the Mongo integration harness, not a live Adyen sandbox.
- Missing outbox event for an expired setup — an expired setup does not currently raise its own outbox event; reviewer confirmed this is not a correctness blocker (no money moves, no subscriber-visible state depends on it) but is worth adding for completeness.

No new correctness concerns identified this round. This closes the one P2 flagged in the previous review.

fahad0208:
Closed both narrow gaps named by the final audit, in commit e3e32616.

**Gap 1 -- frozen provider not shown in client UI**

Added `providerName: string | null` to `SimulatedSubscription` (`client/app/cross-modules/utilities/subscription-simulation/models/subscription-simulation.model.ts`), matching `SubscriptionResponse.ProviderName`'s camelCase JSON shape. Surfaced it in `current-subscription-card.tsx` as a small secondary badge next to the existing status badge, mapped to a friendly label ("Stripe" / "Adyen"), null-safe and hidden when absent -- consistent with how the file already shows other read-only metadata. Checked the other subscription-detail consumers (change-plan/change-quantity/cancel dialogs, overage-terms-section) -- none of them show top-level subscription identity the way the card does, so this was the one place worth touching. Added/updated tests in `current-subscription-card.test.tsx`.

**Gap 2 -- Adyen invoice-breakdown population: found and fixed a real, previously-undiscovered bug**

Traced the actual charge path rather than re-asserting the earlier reasoning, and it turned out to be wrong in one concrete way. `StripeInvoiceBillingGateway` (Stripe's own Invoice API path) has always correctly written the full gross/discount/tax/credit/settlement breakdown onto `PaymentDetail`. But every other provider -- Adyen included -- is charged through `RecurringChargeBillingGateway` -> `IRecurringPaymentService.CreateRecurringPaymentAsync` -> `RecurringPaymentReservationService.CreatePayment`, and `CreateRecurringPaymentRequest` (the request type that path uses) had **no fields at all** for any of this. `RecurringChargeBillingGateway` silently dropped `SubscriptionChargeRequest`'s entire breakdown before it ever reached a `PaymentDetail`. In other words: an Adyen-routed renewal, dunning retry, plan/quantity-change settlement, or usage invoice would have recorded a payment with `SubscriptionGrossAmountMinor`, `SubscriptionNetAmountMinor`, `SubscriptionTaxAmountMinor`, `SubscriptionCreditAmountMinor`, the discount pair, and `SubscriptionSettlement` all left null -- exactly the "silently wrong invoice" failure mode the audit was worried about, not a hypothetical.

Fix: added `SubscriptionInvoiceBreakdown` to `CreateRecurringPaymentRequest` (`Payment.DomainService/Requests/CreateRecurringPaymentRequest.cs`), populated it in `RecurringChargeBillingGateway.ChargeAsync` from `SubscriptionChargeRequest` (mirroring `StripeInvoiceBillingGateway.NewPayment`'s exact "gross > 0" null-vs-zero convention), and copied it onto the created `PaymentDetail` in `RecurringPaymentReservationService.CreatePayment`. Added `server/XUnitTest/Subscription/RecurringChargeBillingGatewayTests.cs` (3 new tests: full breakdown forwarded for an `ADYEN-ONLINE` renewal, an untaxed charge records no tax mode rather than a default, and a settlement charge carries `Settlement` instead of the flat fields) plus 2 new tests in `RecurringPaymentReservationServiceTests.cs` verifying the breakdown lands correctly on the persisted `PaymentDetail`, and the no-breakdown case still records null rather than zeroes for an ordinary non-subscription charge.

**Verification**
- `dotnet build server/Api/Api.csproj --no-incremental`: 0 errors.
- `dotnet test --filter "FullyQualifiedName!~Integration"` (server/): 3414-3415 passed, 4 failed consistently across two full runs -- all four are the known pre-existing `SubscriptionSimulation*` permission-check failures, unrelated to this change. `PeriodicPingBackgroundServiceTests` failed once out of two runs under full-suite load, matching the previously-observed flake; it passed on the second run. All new/changed tests (16 total across the two touched test files) pass consistently.
- `npm run type-check` (client/): 0 errors.
- `npx vitest run current-subscription-card.test.tsx`: 13/13 passed.
- CI on this PR (commit e3e32616): all checks green -- `Analyze (actions/csharp/javascript-typescript)`, `CodeQL`, `initialization`, `repo-hygiene`, both `mongo-integration` jobs, and `validate-client` all passed.

PR #409, opened 2026-09-02
Fix subscription_payment_provider_scope_mismatch on a corrected provider

fahad0208:
## Context

PR #393 ("Merchant-selected Stripe/Adyen provider routing for subscriptions") merged into `dev` while this fix was being investigated, so this follow-up lands as its own PR from the same branch rather than as another commit on #393.

## The real-world failure

A user manually testing #393 configured a real Adyen provider under Payment Providers, selected Adyen as the subscription payment provider on the Merchant Profile page, and then tried to subscribe an organization to a plan. It failed with:

> "The payment provider could not be reached under the merchant configuration this subscription was pinned to" (`subscription_payment_provider_scope_mismatch`)

## Root cause

`BillingAccountRepository.GetOrCreateAndReconcileAsync` keys a billing account's identity on `(TenantId, OrganizationId, ProviderName)`. `BillingAccount.ProviderId`/`ProviderOrganizationId` are written under `$setOnInsert`, so they are set only on the **first** reconcile call for that triple and are otherwise permanently frozen (a deliberate PR #393 review finding, meant to stop a live subscription's charge target from silently moving).

That freeze was too broad: it also permanently pinned an account that never got as far as a confirmed charge. A realistic manual-testing sequence — register Adyen, subscribe (this creates/reconciles the billing account and pins `ProviderId` to whatever readiness resolved), notice a mistake in the Adyen configuration, fix it (which can mean deleting and re-registering the `PaymentProvider` row, minting a new `ItemId`), then subscribe again — leaves the billing account permanently pinned to the first, now-defunct row. Checkout always resolves the provider through the corrected, live row, which forever disagrees with the frozen one, so every subsequent attempt fails closed with `subscription_payment_provider_scope_mismatch` for a merchant that is, in fact, configured correctly.

## The fix

`BillingAccountRepository.ReconcileProviderIdentityAsync` (renamed from `BackfillProviderIdentityAsync`) now re-pins `ProviderId`/`ProviderOrganizationId` to the current readiness resolution on every reconcile, for as long as the account has never taken a real payment (`ProviderCustomerId` unset). The moment `TrySetProviderCustomerAsync` records a confirmed customer, the identity freezes absolutely — exactly the original guarantee, just correctly scoped to when it actually matters. The conditional update is a genuine compare-and-set (filtered on the `ProviderId` value just read, not merely "still unset") so two concurrent reconciles still converge on one winner instead of each reporting a different value.

Two related organization-scope bugs found while tracing the report, both fixed alongside it:

- `SubscriptionCreationService` coerced a genuinely tenant-wide provider's `null` `OrganizationId` into the subscriber's own organization via `readinessResult?.Provider?.OrganizationId ?? context.OrganizationId`, discarding the one fact `BillingAccount.ProviderOrganizationId` exists to record. Now only falls back to `context.OrganizationId` when readiness found no provider at all.
- `SubscriptionCheckoutService`'s `account.ProviderOrganizationId ?? subscription.OrganizationId` fallback couldn't distinguish "never recorded" (legacy account) from "recorded, and genuinely null" (tenant-wide). Both call sites now key off `BillingAccount.ProviderId` (null only for accounts that predate this feature) to make that distinction, via a new `ResolveProviderOrganizationId` helper.

## Testing

- New integration test `A_provider_identity_never_charged_is_repinned_by_reconcile` reproduces the exact reported failure against a real MongoDB: fails on the pre-fix code (`provider-row-misconfigured` stays frozen instead of adopting `provider-row-corrected`), passes after the fix. Verified this directly by reverting just the repository change and re-running.
- `An_already_frozen_provider_identity_is_never_overwritten_by_reconcile` (now `A_provider_identity_already_charged_is_never_overwritten_by_reconcile`) updated to set a real `ProviderCustomerId` first, so it now proves the guarantee that actually needs to hold: frozen once money has moved, not from the first attempt regardless.
- New unit tests in `SubscriptionCreationServiceTests` and `SubscriptionCheckoutServiceTests` cover the tenant-wide-null preservation fix end to end.

Build: `dotnet build server/Api/Api.csproj --no-incremental` — 0 errors.

Tests: `dotnet test --filter "FullyQualifiedName!~Integration"` from `server/` — 3416 passed, 4 known pre-existing `SubscriptionSimulation*` failures (unrelated to this change), 1 skipped. Integration suite (`SubscriptionRepositoryIntegrationTests`, 31 tests) run separately against a local MongoDB — all passed.

Client: `npm run type-check` from `client/` — 0 errors (no-op; this change is server-only).

Commit: 73eb425243e3d0f752c5835520e09a0384e0afe7

### DIFF
--- a/server/Subscription.DomainService/Repositories/BillingAccountRepository.cs
+++ b/server/Subscription.DomainService/Repositories/BillingAccountRepository.cs
@@ -86,53 +86,88 @@ public sealed class BillingAccountRepository : IBillingAccountRepository
                      ?? account;
         }
 
-        return await BackfillProviderIdentityAsync(account, result, cancellationToken);
+        return await ReconcileProviderIdentityAsync(account, result, cancellationToken);
     }
 
     /// <summary>
-    /// One-time self-healing backfill of <see cref="BillingAccount.ProviderId"/> and
-    /// <see cref="BillingAccount.ProviderOrganizationId"/> onto a legacy account that predates
-    /// this PR's provider-identity work and so was created with both left null.
+    /// Keeps <see cref="BillingAccount.ProviderId"/> and
+    /// <see cref="BillingAccount.ProviderOrganizationId"/> pointed at whatever readiness resolved
+    /// just now, for as long as doing so is still safe -- and freezes them the moment it stops
+    /// being safe.
     /// </summary>
     /// <remarks>
+    /// An account is one per organization and provider and outlives every subscription on it (see
+    /// <see cref="IBillingAccountRepository.GetOrCreateAndReconcileAsync"/>), so the very first
+    /// signup attempt for a given (tenant, organization, provider name) is not necessarily the one
+    /// that ends up paying anything: it can be misconfigured, abandoned, or superseded by a
+    /// corrected <c>PaymentProvider</c> row before a card is ever actually charged or stored. Only
+    /// once <see cref="BillingAccount.ProviderCustomerId"/> is set does that account name a real
+    /// customer at a real provider -- filled in from a confirmed charge, never guessed at (see its
+    /// own remarks) -- and only from that point on would moving <see cref="BillingAccount.ProviderId"/>
+    /// risk stranding a saved card or repointing a live mandate. Before then, re-pinning to the
+    /// current resolution is not merely safe, it is the only way an operator who fixes a
+    /// misconfigured provider between two subscribe attempts is not stuck forever with the first,
+    /// broken one -- see PR #393's <c>subscription_payment_provider_scope_mismatch</c> finding.
+    /// <para>
     /// Deliberately not folded into <see cref="Reconciliation"/>: that single find-and-modify
-    /// cannot express "write this field only if it is currently null" — <c>$set</c> and
-    /// <c>$setOnInsert</c> both act unconditionally on their own side of insert-vs-update — so the
-    /// conditional write here follows this codebase's established compare-and-set convention
+    /// cannot express "write this field only if the account has taken no money yet" — <c>$set</c>
+    /// and <c>$setOnInsert</c> both act unconditionally on their own side of insert-vs-update — so
+    /// the conditional write here follows this codebase's established compare-and-set convention
     /// instead (see <c>PaymentRepository.TryRecordSetupTokenConfirmedAsync</c> in
-    /// Payment.DomainService): a conditional update filtered on the field still being null.
-    /// That filter is what makes
-    /// this strictly additive and one-directional — a billing account's provider identity is
-    /// frozen once set, and this must never be the thing that silently moves it, only the thing
-    /// that fills in a value legacy accounts were never given a chance to record.
+    /// Payment.DomainService): a conditional update filtered on <c>ProviderCustomerId</c> still
+    /// being unset. That filter is what makes this reversible only until real money is on the
+    /// line, and absolute afterwards -- a billing account that has already taken a payment must
+    /// never have its provider identity moved by anything, this method included.
+    /// </para>
     /// <para>
     /// Uses <c>FindOneAndUpdate</c> with <see cref="ReturnDocument.After"/> rather than an
     /// <c>UpdateOneAsync</c> read off its <c>ModifiedCount</c>, so a losing writer in a race
-    /// between two concurrent backfills of the same legacy account still gets back whatever is
+    /// between two concurrent reconciliations of the same account still gets back whatever is
     /// actually on the document now -- the value the winner wrote -- instead of the stale
-    /// pre-update <paramref name="stored"/> it was handed, which would still show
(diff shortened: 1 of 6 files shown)


fahad0208:
## Update: PR #393 merged mid-investigation

While tracing this, PR #393 merged into `dev` (2026-09-02T21:30:59Z) before the fix commit could be pushed to it, and the branch became a protected/closed ref that no longer accepts CI-triggering pushes. This PR (#409), opened from the same `feat/merchant-payment-provider-selection` branch against `dev`, carries the one follow-up commit and is where CI actually ran.

## The real-world failure

A user manually testing #393 configured a real Adyen provider under Payment Providers, selected Adyen as the subscription payment provider on the Merchant Profile page, and then tried to subscribe an organization to a plan. It failed with:

> "The payment provider could not be reached under the merchant configuration this subscription was pinned to" (`subscription_payment_provider_scope_mismatch`)

## Root cause

`BillingAccountRepository.GetOrCreateAndReconcileAsync` keys a billing account's identity on `(TenantId, OrganizationId, ProviderName)`. `BillingAccount.ProviderId` / `ProviderOrganizationId` are written under Mongo's `$setOnInsert`, so they are only ever set on the **first** reconcile for that triple and are otherwise frozen forever — a deliberate #393 review finding meant to stop a *live* subscription's charge target from silently moving.

That freeze was too broad: it also permanently pinned an account that never got as far as a confirmed charge. The realistic manual-testing sequence that reproduces it: register Adyen, subscribe (this reconciles the billing account and pins `ProviderId` to whatever readiness resolved at that moment), notice a mistake in the Adyen config, fix it (which can mean deleting and re-registering the `PaymentProvider` row, minting a new `ItemId`), then subscribe again. The billing account stays pinned to the first, now-defunct row forever. Checkout always resolves the provider through the corrected, live row, which permanently disagrees with the frozen one, so every retry fails closed with `subscription_payment_provider_scope_mismatch` for a merchant that is, in fact, configured correctly.

## The fix

`BillingAccountRepository.ReconcileProviderIdentityAsync` (renamed from `BackfillProviderIdentityAsync`) now re-pins `ProviderId`/`ProviderOrganizationId` to the current readiness resolution on every reconcile, for as long as the account has never taken a real payment (`ProviderCustomerId` unset). The moment `TrySetProviderCustomerAsync` records a confirmed customer, the identity freezes absolutely — the original guarantee, correctly scoped to when it actually matters. The conditional update is a genuine compare-and-set (filtered on the exact `ProviderId` value just read, not merely "still unset"), so two concurrent reconciles still converge on one winner instead of each reporting a different value.

Two related organization-scope bugs found while tracing the report, fixed alongside it:
- `SubscriptionCreationService` coerced a genuinely tenant-wide provider's `null` `OrganizationId` into the subscriber's own organization via `readinessResult?.Provider?.OrganizationId ?? context.OrganizationId`, discarding the one fact `BillingAccount.ProviderOrganizationId` exists to record. Now only falls back to `context.OrganizationId` when readiness found no provider at all.
- `SubscriptionCheckoutService`'s `account.ProviderOrganizationId ?? subscription.OrganizationId` fallback couldn't distinguish "never recorded" (legacy account) from "recorded, and genuinely null" (tenant-wide). Both call sites now key off `BillingAccount.ProviderId` (null only for accounts that predate this feature) via a new `ResolveProviderOrganizationId` helper.

## Testing

- New integration test `A_provider_identity_never_charged_is_repinned_by_reconcile` reproduces the exact failure against a real MongoDB — verified it fails on the pre-fix code (stays pinned to `provider-row-misconfigured` instead of adopting `provider-row-corrected`) and passes after the fix, by reverting just the repository change and re-running.
- `An_already_frozen_provider_identity_is_never_overwritten_by_reconcile` renamed to `A_provider_identity_already_charged_is_never_overwritten_by_reconcile` and updated to record a real `ProviderCustomerId` first, so it now proves the guarantee that actually needs to hold.
- New unit tests in `SubscriptionCreationServiceTests` and `SubscriptionCheckoutServiceTests` cover the tenant-wide-null preservation fix end to end.

## Commit

`73eb425243e3d0f752c5835520e09a0384e0afe7`

## Results — all green

- **Build**: `dotnet build server/Api/Api.csproj --no-incremental` — 0 errors.
- **Tests**: `dotnet test --filter "FullyQualifiedName!~Integration"` from `server/` — 3416 passed, 4 known pre-existing `SubscriptionSimulation*` failures (unrelated), 1 skipped. `SubscriptionRepositoryIntegrationTests` (31 tests, incl. the new repro) run separately against a local MongoDB — all passed.
- **Client type-check**: `npm run type-check` from `client/` — 0 errors (no-op; server-only change).
- **CI on this PR**: `Analyze (actions)`, `Analyze (csharp)`, `Analyze (javascript-typescript)`, `CodeQL`, `initialization`, `mongo-integration`, `repo-hygiene` — all passing.

PR #452, opened 2026-09-05
fix(subscription): stop a lost payment webhook from stranding a subscriber

fahad0208:
## Summary

During a billing simulation, a subscriber completed a CHF 0.00 card authorisation at Stripe and
closed the tab. Their subscription stayed `Incomplete` forever. The immediate trigger (a dev
webhook-subscription gap) is being fixed separately; this PR fixes why a missing/late webhook was
so expensive to begin with. Three compounding defects, all in `SubscriptionActivationProcessor` /
`SubscriptionCheckoutService`:

1. **Undecided treated as failure.** Once `ActivationMaxAttempts` was exhausted, the sweep
   abandoned the link unconditionally — even though `ClassifyAsync` already knows the difference
   between a decided failure and a payment nobody has said anything about yet.
2. **The safety net excluded exactly what it was handed.** `RecoverStaleAsync` only ever looked at
   subscriptions with *no* payment link, so an `Abandoned` card-setup link — which `AbandonAsync`
   deliberately leaves the subscription `Incomplete` for, on the promise this sweep would finish or
   end it — was never looked at again.
3. **Retrying an abandoned setup replayed the dead session.** `StartPaymentMethodSetupAsync` fell
   through to `StartCardSetupAsync` for an `Abandoned` link, which derives its idempotency key from
   an unchanged attempt number — so Stripe replayed the same expired `cs_test_…` session instead of
   opening a new one.

## What changed

- **`SubscriptionActivationProcessor`** (`RescheduleAsync`): once the retry budget is exhausted, a
  new optional `ISubscriptionPaymentReconciler` is asked what the provider actually decided, before
  anything is abandoned. A provider-confirmed activation or failure settles the link exactly as an
  on-time webhook would; anything short of a decided answer (unreachable provider, no reconciler,
  session still open) leaves the link `Pending` on a wider, dedicated cadence
  (`ActivationProviderConfirmationRetrySeconds`) rather than declaring failure on silence.
- **`ClassifyAsync`**: `PaymentStatuses.Expired` is now a decided failure. It wasn't, so a card
  setup expired by `PaymentMethodSetupExpiryProcessor`'s own sweep was classified `Undecided`
  forever instead of `Abandoned`.
- **`RecoverStaleAsync`** (B2): now also reconciles `Abandoned` links (previously skipped
  unconditionally, same as `Pending`/`Applied`) — a late confirmation activates the subscription,
  otherwise the subscription is finally expired instead of sitting `Incomplete` indefinitely.
- **`SubscriptionCheckoutService.StartPaymentMethodSetupAsync`** (B3): an `Abandoned` setup link
  now routes to the existing `RetryCardSetupAsync` (which bumps the attempt through its own
  compare-and-set) instead of falling through to `StartCardSetupAsync`'s unchanged key.
- **New in `Payment.DomainService`: `StripeCheckoutReconciliationService`** (`ISubscriptionPaymentReconciler`).
  Reads the Checkout Session with its intent's `payment_method` expanded (`expand[]=payment_intent.payment_method`,
  `expand[]=setup_intent.payment_method`) — needed because completing a card-setup activation
  requires the actual saved card, which the plain session read never returns — and applies the
  outcome through the *same* `IPaymentWebhookStateTransitionService` /
  `IPaymentMethodSetupWebhookStateTransitionService` a genuine webhook uses, so a later, truly late
  webhook for the same event is a harmless idempotent no-op rather than a double application.
  Stripe-only; a provider without a registered reconciler is treated as "cannot observe" and never
  turns into an abandonment. **Not verified against a live Stripe sandbox** — no sandbox available
  in this environment — covered instead by unit tests against mocked HTTP responses.

## Why this is safe

- `ISubscriptionPaymentReconciler` is optional (nullable, DI-resolved) everywhere it's consumed —
  a deployment without it, or a provider it can't observe, behaves exactly like "no answer yet",
  never like "declare failure".
- The reconciler never invents its own confirmation logic: it constructs a `PaymentWebhookInbox`
  from the provider read and hands it to the real transition services, so every existing
  invariant (webhook-confirmed-at gate, amount/currency check, outbox dedup key, the setup's
  two-signal completion) applies unchanged.
- `RescheduleAsync` now returns whether a provider read actually settled the link, so the sweep's
  own `settled` count and audit trail stay accurate instead of just assuming "deferred".

## Test plan

- [x] `dotnet build server/Blocks.slnx` — 0 errors (same warning baseline as `dev`)
- [x] `dotnet test server/XUnitTest/XUnitTest.csproj --filter "FullyQualifiedName!~Integration"` —
      3680 passed, 4 pre-existing failures unrelated to this change (confirmed present on `dev`
      before this branch: `SubscriptionSimulationServiceTests` /
      `SubscriptionSimulationDataConsoleServiceTests`), 1 skipped
- [x] New/updated unit tests: `SubscriptionActivationProcessorTests` (budget-exhausted +
      provider-confirmed activation/failure, no-decided-answer holds open, expired-payment
      classification, abandoned-setup-link recovery with/without confirmation, pending-link still
      skipped), `SubscriptionCheckoutServiceTests` (abandoned setup retried under a new key),
      `StripeCheckoutReconciliationServiceTests` (unsupported provider, missing session id, unsafe
      endpoint, still-open session, session-id mismatch, completed charge, completed card setup —
      the literal incident case, expired session, malformed synthetic event never throws)
- [ ] e2e / manual dev verification (webhook-gap independent, per the four end-to-end scenarios
      in the originating plan) — not run in this environment

### DIFF
--- a/server/Payment.DomainService/Providers/Stripe/StripeCheckoutSessionReconciliation.cs
+++ b/server/Payment.DomainService/Providers/Stripe/StripeCheckoutSessionReconciliation.cs
@@ -1,0 +1,91 @@
+using System.Text.Json.Serialization;
+
+namespace Payment.DomainService.Providers.Stripe;
+
+/// <summary>
+/// A Checkout Session read with <c>payment_intent.payment_method</c> and
+/// <c>setup_intent.payment_method</c> expanded, so a lost webhook's outcome can be reconstructed
+/// from one call instead of guessed at from the bare session Standard reads use.
+/// </summary>
+/// <remarks>
+/// Deliberately a separate model from <see cref="StripeCheckoutSession"/> rather than adding these
+/// fields to it. That type's <c>PaymentIntent</c> is a plain id string everywhere else it is read,
+/// and Stripe returns a full object in that same field once expansion is requested — modelling
+/// both shapes on one property would make every existing caller's assumption silently wrong the
+/// day this type's query parameters changed.
+/// </remarks>
+public sealed class StripeCheckoutSessionReconciliation
+{
+    [JsonPropertyName("id")]
+    public string? Id { get; set; }
+
+    /// <summary><c>open</c>, <c>complete</c> or <c>expired</c>.</summary>
+    [JsonPropertyName("status")]
+    public string? Status { get; set; }
+
+    /// <summary><c>unpaid</c>, <c>paid</c> or <c>no_payment_required</c>.</summary>
+    [JsonPropertyName("payment_status")]
+    public string? PaymentStatus { get; set; }
+
+    [JsonPropertyName("client_reference_id")]
+    public string? ClientReferenceId { get; set; }
+
+    [JsonPropertyName("payment_intent")]
+    public StripeReconciliationIntent? PaymentIntent { get; set; }
+
+    [JsonPropertyName("setup_intent")]
+    public StripeReconciliationIntent? SetupIntent { get; set; }
+
+    [JsonPropertyName("amount_total")]
+    public long? AmountTotal { get; set; }
+
+    [JsonPropertyName("currency")]
+    public string? Currency { get; set; }
+
+    /// <summary>Present instead of the session fields when Stripe rejects the call.</summary>
+    [JsonPropertyName("error")]
+    public StripeError? Error { get; set; }
+}
+
+/// <summary>A PaymentIntent or SetupIntent, expanded far enough to see what it stored.</summary>
+public sealed class StripeReconciliationIntent
+{
+    [JsonPropertyName("id")]
+    public string? Id { get; set; }
+
+    /// <summary>
+    /// PaymentIntent: <c>succeeded</c>, <c>requires_payment_method</c>, <c>canceled</c>, and so on.
+    /// SetupIntent: <c>succeeded</c>, <c>requires_payment_method</c>, <c>canceled</c>.
+    /// </summary>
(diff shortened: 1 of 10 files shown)


fahad0208:
Added a follow-up commit that closes a gap the first commit introduced: a budget-exhausted, undecided activation link (B1) had no way to ever reach a terminal state for a provider with no registered `ISubscriptionPaymentReconciler` (Adyen today) — it would poll `Pending` forever and the subscription would sit `Incomplete` permanently, which is the exact failure mode `AbandonAsync` warns against, just reached from the age-based sweep's side.

- New `SubscriptionOptions.ActivationUnresolvedCeilingHours` (default 48h) — deliberately far past `InitialChargeGraceMinutes` (60min) so a subscriber still inside a hosted checkout session's own ~24h lifetime is never expired by this.
- `RecoverStaleAsync` now also handles budget-exhausted `Pending` links past this ceiling: one final provider read, and if that still can't decide, the link is settled `Abandoned` and the subscription expired.
- 6 new unit tests covering: within-ceiling (must be left alone — the regression guard), past-ceiling-undecided (ends it), past-ceiling-now-confirmed (activates instead), and `Applied` links (still skipped).

Build: 0 errors. Full non-integration suite: 3684 passed, same 4 pre-existing unrelated failures as before (`SubscriptionSimulation*Tests`, confirmed present on `dev` prior to this branch), 1 skipped.

PR #456, opened 2026-09-06
Revert [ProtectedEndPoint] on subscription and payment endpoints

fahad0208:
Reverts cb17b8ba and ff652314. Restores `[Authorize]` on every subscription and payment endpoint, and the `SubscriptionBackgroundWorkOperator` policy in `Program.cs`.

## Why

`[ProtectedEndPoint("<resource>")]` asks the framework whether the caller holds a named resource in the tenant's `Permissions` collection. **Those rows do not exist yet**, so `ProtectedEndpointAccessHandler` correctly refuses everyone and all 68 endpoints answer 403.

The change was merged (#453) and promoted to `stg` and `main` before the permissions it depends on were seeded. Authentication-only authorization is the weaker check, but it works — and a working weaker check beats a correct check that denies every caller.

## Scope

Every file the original change touched is byte-identical to its state before it — verified with `git diff cb17b8ba~1 -- <touched files>`, which is empty.

- 68 endpoints across 13 controllers: `[ProtectedEndPoint(...)]` → `[Authorize]`
- `SubscriptionBackgroundWorkOperator` policy restored in `Program.cs`
- `server/Api/Controllers/README.md` removed
- `SubscriptionSimulationGuard` restored to taking `callerPermissions`

## Known: 4 tests go red again

The revert restores `SubscriptionSimulationGuard`'s permission arm and its tests:

- `SubscriptionSimulationGuardTests.Refuses_the_console_without_the_permission`
- `SubscriptionSimulationGuardTests.Refuses_the_console_with_no_permissions_at_all`
- `SubscriptionSimulationServiceTests.Refuses_the_console_without_the_simulation_permission`
- `SubscriptionSimulationDataConsoleServiceTests.Find_refuses_the_console_without_the_simulation_permission`

These were **already failing on `dev` before #453** — the guard's permission check is commented out in the restored code — and `dev` was promoted to `stg` and `main` with them red. They are not what broke the endpoints, so they are left exactly as the revert produces rather than half-reverted. Worth fixing separately.

## Verification

- `dotnet build Blocks.slnx` — 0 errors.
- `dotnet test --filter "FullyQualifiedName!~Integration"` — 3685 passed, 4 failed (the pre-existing ones above).

## To re-apply later

1. Seed the 30 resources into every tenant's `Permissions` collection — `scripts/blocks-utilities-permissions/` on `feat/subscription-payment-protected-endpoints`, which is not merged and still has the seeder, wrappers and runbook.
2. Confirm a real token authorises against them in dev.
3. Revert this commit.


fahad0208:
Not reverting `dev` — the change stays there deliberately. Production is being unblocked with a `main`-only hotfix instead: #457.

PR #506, opened 2026-09-16
fix(subscription): stop pre-migration aggregate rows appearing as per-user usage rows

fahad0208:
## Summary
`GET /subscription-usage/current` was returning every meter twice for any subscription carrying at least one usage document written before the per-user-usage-breakdown feature shipped (`SchemaVersion < 3` — no `UserId` field in its BSON at all, not even `""`).

## Root cause
`ListUserRowsAsync` and `ListUserRowsBehindAsync` (`SubscriptionUsageCurrentRepository.cs`) both filtered with `Builders<T>.Filter.Ne(current => current.UserId, string.Empty)`. MongoDB's `$ne` matches a document where the field is **missing entirely**, not only one where it holds something other than `""`. So every pre-migration aggregate row — which has no `UserId` element at all — was returned by both methods as if it were a genuine per-user row. On deserialization the missing field defaults to `string.Empty` (the C# property initializer), making it indistinguishable from the real aggregate row already in the response.

The response is built as `aggregate items ++ user items` (`UsageRecordingService.ReadOf`), so the aggregate row ended up counted in both halves — every meter appearing twice, both copies showing `"userId": ""`.

One visible tell in the bug report: the two `LIFETIME`-period copies of each meter serialized `periodEndUtc` differently (with vs without a trailing `Z`) — one came straight from `MeterPeriodResolver` in memory (`DateTime.MaxValue`, `Kind.Unspecified`), the other from a Mongo round-trip (the driver marks deserialized `DateTime`s `Kind.Utc`). Same instant, two `Kind` tags, two JSON forms — not a separate bug, just what made the duplication visible.

## Fix
- Added a shared `UserRowFilter()` helper requiring the field to **exist** as well as differ from `""`, used by both `ListUserRowsAsync` and `ListUserRowsBehindAsync` (they shared the same broken predicate shape, so one fix covers both — `ListUserRowsBehindAsync` feeds the repair sweep and had the identical flaw, which would have made it misidentify legacy aggregate rows as "per-user rows behind" and attempt needless repairs against them).
- Normalized the `Never`-reset meter's lifetime period end to `Kind.Utc` in `MeterPeriodResolver`, so the same instant no longer serializes two different ways depending on which code path produced it.

## Verification
- Two new integration tests directly proving the fix at the repository level (`Listing_user_rows_excludes_a_pre_migration_document_with_no_user_id_field`, `Listing_user_rows_behind_excludes_a_pre_migration_document_with_no_user_id_field`), inserting a raw BSON document with no `UserId` field (mirroring the existing `A_republish_backfills_the_missing_user_id_onto_a_pre_migration_document` test, whose own remark already predicted this exact bug) and asserting both listing methods now return nothing for it.
- `dotnet build server/Blocks.slnx` — 0 errors.
- `UsageRecordingServiceTests` (mocked, unaffected by the repository-level filter change) — 42/42 passing.
- Ran the Mongo integration suite against a real MongoDB (via `BLOCKS_IT_MONGO`) — 33/33 passing, including the two new tests above.

## Test plan
- [x] `dotnet build server/Blocks.slnx` — 0 errors
- [x] `dotnet test --filter "FullyQualifiedName~UsageRecordingServiceTests"` — 42/42 passing
- [x] `dotnet test --filter "FullyQualifiedName~SubscriptionUsageCurrentIntegrationTests"` against a real MongoDB (`BLOCKS_IT_MONGO`) — 33/33 passing
- [ ] Manual smoke test: call `GET /subscription-usage/current` for a subscription with a pre-migration usage document and confirm each meter appears once

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


fahad0208:
Ran the Mongo integration suite against a real MongoDB instance (via \`BLOCKS_IT_MONGO\`, a throwaway per-run database that gets dropped after — never a persisted one):

\`\`\`
dotnet test --filter "FullyQualifiedName~SubscriptionUsageCurrentIntegrationTests"
Passed!  - Failed: 0, Passed: 33, Skipped: 0, Total: 33
\`\`\`

All 33 pass, including the two new tests added in this PR (\`Listing_user_rows_excludes_a_pre_migration_document_with_no_user_id_field\`, \`Listing_user_rows_behind_excludes_a_pre_migration_document_with_no_user_id_field\`), confirming the fix against a real database rather than only the mocked service tests. Updating the test plan checklist accordingly.

PR #520, opened 2026-09-17
feat(simulation): add a discount code tester to the simulation console

fahad0208:
## Why

The simulation screen could carry a discount code only inside the subscribe dialog — on the way to
a real subscription, and only for an organization that did not already have one. There was no way
to ask the questions an integrator actually asks about a code: is it live, what is it worth against
this price, and why was it refused?

## What

A **Discount code** card on the simulation page (below the plan catalogue): pick a plan and price,
enter a code, hit **Test code**, and read back the server's own verdict.

- Wires up `POST /api/subscription-discounts/preview` — the buyer-facing, read-only endpoint that
  already existed (`SubscriptionDiscountPreviewController` → `PreviewDiscountAsync`) and was unused
  by the client. No new backend API was needed, so none was added.
- Shows the verdict (`Applied`, `NotFound`, `NotStarted`, `Expired`, `NotApplicable`,
  `AlreadyRedeemed`, `Unavailable`), the server's message, and the **reason code** an integration
  would have to handle — the code is what distinguishes the outcomes, and it is not reconstructible
  from the sentence.
- Shows the resulting quote (due now + next renewal), the campaign note when the code is a platform
  campaign, and any blocker that would stop a real signup without touching the code's own verdict.
- A rejection is data, not an error: the standard undiscounted quote comes back alongside the
  reason, so the two prices can be read side by side.
- Nothing is charged and no redemption is reserved — a campaign limited to a fixed number of
  redemptions is not spent by being tested.
- Any edit to the plan, price or code discards the verdict it produced: a code applied to one price
  says nothing about another.

`MoneyBreakdown` moved out of `subscribe-dialog.tsx` into its own component rather than being
duplicated for the second quote that now reads the same way — no behaviour change there.

## Checks

- `tsc --noEmit` — clean
- `eslint` on the touched module — clean
- `vitest` on `subscription-simulation` — 9 files, 93 tests pass (6 of them new for the card:
  empty-code guard, request shape, applied verdict, rejection with reason code, verdict discarded
  on edit, failed check)


fahad0208:
Pushed a second commit: the quote now shows the charge **actually taken next** whenever it is not the recurring price.

`nextCharge` was broken out only when `prorated` was true, on the assumption that an un-prorated next charge is the same figure as `nextRenewal`. The response in the report shows it is not — same full period, same date, nothing prorated, two different amounts:

- `nextRenewal` prices a full period against the discount periods applied **today** → CHF 972.90
- `nextCharge` projects the consumption the opening payment will have made → CHF 1,026.95 (after #521)

A code limited to one period is spent by the time that charge falls due, so the dialog was showing only the lower figure, labelled "Next renewal" with the date the higher one is actually charged.

Now: one helper (`nextChargeDiffersFromRenewal`) compares the amounts rather than reading `prorated` alone, used by both the subscribe dialog and the new discount tester. When they disagree, the real charge renders as "Charged next" with its date and the full-period figure is relabeled "Recurring price"; when they agree, nothing changes and one figure shows as before. The model comment that claimed "the amount alone cannot" tell the cases apart is corrected.

Also synced one existing test fixture that overrode `nextRenewal` with tax while leaving `nextCharge` untaxed — an inconsistency no server returns for a single un-prorated period, and it would otherwise have been broken out as a differing charge.

Checks: `tsc --noEmit` clean, eslint clean, `vitest` on `subscription-simulation` 97/97 (4 new).

